PR #16, opened 2018-06-27
Updating PropTypes for Section and PageHeaderTitle

ryelo:
This should allow for the first stable release of the component library

#15 
#14


ryelo:
@karelhala What's the best way to do that? Do I need to update the `package.json` or just add a commit with `<x.x.x>`?

PR #20, opened 2018-06-27
Fixes/package info

ryelo:
Adding a few fields to the package.

npm was giving warnings when `npm install`


README.md
```diff
@@ -1,177 +1,37 @@
...
+# Development
+Create a component inside of the PresentationalComponent folder
 
-### Testing
-- Travis is used to test the build for this code.
-    - `npm test` will run the test
-
-# Documentation links
- * Utils
-   * [Redux](doc/utils/redux.md)
```
karelhala: We need to document utils and maybe even components, how about creating new section with doc links to each util/component?

ryelo: Updated!

PR #24, opened 2018-07-17
Introduce simple filter with button and text input <?.x.?>

karelhala:
### New filter component
This PR introduces new component called `SimpleTableFilter` for filtering in tables
 * input -> onChange has callback `onFilterChange` to filter without need of pressing button (not good for server generated reports)
 * button -> onClick has callback `onButtonClick` this callback will get active filter

#### Props
```javascript
{
  buttonTitle: PropTypes.string,
  placeholder: PropTypes.string,
  className: PropTypes.string,
  onButtonClick: PropTypes.func,
  onFilterChange: PropTypes.func
}
```

#### Example
screenshot from 2018-07-17 15-27-12

### DIFF
--- a/src/PresentationalComponents/SimpleTableFilter/SimpleTableFilter.js
+++ b/src/PresentationalComponents/SimpleTableFilter/SimpleTableFilter.js
@@ -1,0 +1,50 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import { Button, ButtonVariant } from '@patternfly/react-core';
+
+class SimpleFilter extends Component {
+  constructor(props) {
+    super(props);
+    this.onInputChange = this.onInputChange.bind(this);
+    this.onFilterSubmit = this.onFilterSubmit.bind(this);
+    this.state = {
+      activeFilter: ''
+    }
+  }
+
+  onInputChange(event) {
+    this.setState({ activeFilter: event.target.value });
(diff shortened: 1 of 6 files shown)


src/PresentationalComponents/SimpleTableFilter/SimpleTableFilter.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+SimpleFilter.propTypes = {
+  buttonTitle: PropTypes.string,
+  placeholder: PropTypes.string,
+  className: PropTypes.string,
+  onButtonClick: PropTypes.func,
+  onFilterChange: PropTypes.func
+}
```
ryelo: Do we need to assign default values to any of these propTypes or require anything?

ryelo: Do we want to require any of these? Couldn't someone just clear out the placeholder string or buttonTitle and have blank spaces?

karelhala: These proptypes are for users of this component so they know what props and what values to pass into. Of course if somebody doesn't want to have placeholder he can pass empty string, if somebody doesn't want to have button they can just pass ampty string into `buttonTitle` and no button will be visible.

PR #26, opened 2018-07-20
Introduce Input component as wrapper for styling <?.x.?>

karelhala:
This is just a wrapper for input components so we don't have to decide which class to use for which input.

### DIFF
--- a/src/PresentationalComponents/Input/Input.js
+++ b/src/PresentationalComponents/Input/Input.js
@@ -1,0 +1,24 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import classnames from 'classnames';
+
+const checkTypes = ['checkbox', 'radio'];
+
+const Input = ({type = 'text', ariaLabel = type, className, ...props}) => {
+  const classes = checkTypes.indexOf(type) !== -1 ? 'pf-c-check' : 'pf-c-form-control';
+  return (
+    <input {...props}
+      type={type}
+      aria-label={ariaLabel}
+      className={classnames(classes, className)}
+    />
+  );
+};
+
+Input.propTypes = {
+  type: PropTypes.string,
+  className: PropTypes.string,
+  ariaLabel: PropTypes.string
+}
+
+export default Input;
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,6 @@ yarn-error.log*
 coverage
 .storybook
 
+demo/
+
 .DS_Store
--- a/doc/components/input.md
+++ b/doc/components/input.md
@@ -1,0 +1,23 @@
+# Input
+This component will help when using input with any type you want. You don't have to worry about applied classes, just 
+pass prop `type` and this component will decide which classes to apply.
+
+### Usage
+Import Input component and it's styles, pass which type you want to use (defaults to `text`) and use as you'd use basic
+HTML `input`.
+```javascript
+import { Input } from '@redhat-insights/insights-frontend-components/components/Input';
+import '@redhat-insights/insights-frontend-components/components/Input.css'
+import React from 'react';
+
+const MyCmp = () => (
+  <div>
+    <Input type="text" />
+    <Input type="radio" />
+    <Input type="check" />
+  </div>
+)
+
+```
+
+### Props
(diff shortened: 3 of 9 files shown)


src/PresentationalComponents/Input/Input.js
```diff
@@ -0,0 +1,22 @@
...
+import React from 'react';
+import classnames from 'classnames';
+
+const checkTypes = ['checkbox', 'radio'];
+
+const Input = ({type, className, ...props}) => {
+  const classes = checkTypes.hasOwnProperty(type) ? 'pf-c-check' : 'pf-c-form-control';
+  return (
+    <input {...props}
+      type={type}
```
ryelo: Can we default `<Input>` with no `type` to a text field?

src/PresentationalComponents/Input/Input.js
```diff
@@ -0,0 +1,22 @@
...
+      className={classnames(classes, className)}
+    />
+  );
+};
+
+Input.propTypes = {
+  type: PropTypes.string,
+  className: PropTypes.string
+}
```
ryelo: Are we going to be adding the ability to have disabled inputs?

src/PresentationalComponents/Input/Input.js
```diff
@@ -0,0 +1,22 @@
...
+
+const checkTypes = ['checkbox', 'radio'];
+
+const Input = ({type, className, ...props}) => {
+  const classes = checkTypes.hasOwnProperty(type) ? 'pf-c-check' : 'pf-c-form-control';
+  return (
+    <input {...props}
+      type={type}
+      className={classnames(classes, className)}
+    />
```
ryelo: These really should be paired with labels for accessibility
```HTML
<input id="foo" type="checkbox" value={children}/>
<label for="foo">{children}</label>
```

ryelo: Needs to assign `checked` attribute for any checked elements in radios and checkboxes

ryelo: This should also apply for `type='text'`. There should never be an input without a label, even if using a placeholder.

src/PresentationalComponents/Input/Input.js
karelhala: Yes, if you look at [props spread](https://github.com/RedHatInsights/insights-frontend-components/pull/26/files#diff-374245968821659edef31240ab6fd102R10) disabled prop will be passed into input.

src/PresentationalComponents/Input/Input.js
karelhala: With this you should be able to do something like
```JSX
import React from 'react';
import { Input } from '@redhat-insights/insights-frontend-components';

const yourCmp = () => (
  <label>Foo</label>
  <Input />
);
```

But sure labeled Input could be exported component as well, so person using this will have option to use input with label as easy as possible.

karelhala: `Checked` attribute will be applied from `{...props}` spread so if we don't need to do anything with chacked no need to add it to props.

karelhala: Oh I see, mhhh... well I guess we can provide two components labeled input (mostly aplicable to checkboxes and radio inputs) and regular input. And if consumers want to use checkbox without label it's up to them.

karelhala: @ryelo last changes to this are that when using `Input` component consumer should set `aria-label`  if not it will default to `type`. So `<Input type="checkox" />` will look like this in HTML
```HTML
<input type="checkbox" aria-label="checkbox" class="pf-c-check">
```

ryelo: The first example is correct. Otherwise, it would need to look like
```HTML
<input type="checkbox" id={props.foo} name={props.foo}>
<label for={props.foo}>{children}</label>
```

karelhala: Makes sence, okay it should be fine now:
```HTML
<input type="checkbox" class="pf-c-check" aria-label="checkbox">
<label class="pf-c-form__label" for="15326125380690.a5cf3ng4qe9">
  <input name="15326125380690.a5cf3ng4qe9" id="15326125380690.a5cf3ng4qe9" type="checkbox" class="pf-c-check"> 
  Some label
</label>
```
Those IDs are autogenerated from time + math random.

PR #29, opened 2018-07-30
Fix bug when page is changed from page input <?.?.x>

karelhala:
When changing page with page input call `onSetPage` callback.

Add Table docs


ryelo:
Sweet. Looks great!

Issue #30, opened 2018-08-06
Component: ins-battery


ryelo:
![image](https://user-images.githubusercontent.com/12520842/43727352-b23a2ba4-996f-11e8-8d10-fe04936379a3.png)

PR #31, opened 2018-08-06
Adding battery component

ryelo:
#30


ryelo:
Eventually, the battery can be on its own with a tooltip component

ryelo:
#34

src/PresentationalComponents/Battery/battery.js
```diff
@@ -0,0 +1,47 @@
...
+import './battery.scss';
+
+/**
+ * This is a dumb component that only recieves properties from a smart component.
+ * Dumb components are usually functions and not classes.
+ *
+ * @param props the props given by the smart component.
+ */
+
+class Battery extends React.Component {
```
karelhala: No need to use class in here since there are no methods used with lifecycle of component so it might look better using inline function.

Also consider allowing consumer to pass additional `className` and spread `props` in there so consumers can for instance pass their `styles` object, `disabled` and other attributes.
```JSX
const Battery = ({severity, label, labelHidden, className, ...props}) => {
  let severityClass ...
  if ...
}
```

src/PresentationalComponents/Battery/index.js
```diff
@@ -0,0 +1 @@
+export {default as Battery } from './battery';
```
karelhala: This is great way of exporting components! Please add `export * from './PresentationalComponents/Battery';` into `src/index.js` as well so consumers can import this component either calling `import { Battery } from '@red-hat-insights/insights-frontend-components/components/Battery';` or simply `import { Battery } from '@red-hat-insights/insights-frontend-components';`

PR #38, opened 2018-08-08
Add callbacks for cell and row clicks in Table

karelhala:
No UI changes.

Also allow consumers to pass object instead of array as rows so it's easier for them to look trough items (clicking on row, selected, etc.)


karelhala:
CC @jiridostal

PR #41, opened 2018-08-09
Introduce new folder with charts, add new chart matrix with cirlces as elements

karelhala:
### New things
 * Component - Matrix: represents 4 quadrant with circles in them
 * Folder layout - charts: contains chart components with settings in build process to publish these components for consumers

### Test data:
```JS
const chartData = {
  topRight: [
    {position: [2, 3], label: 'Something'}
  ],
  topLeft: [{position: [1, 1], label: 'One'},{position: [1, 1], label: 'Two'}],
  bottomRight: [{position: [0, 1], label: 'Another'}],
  bottomLeft: [
    {position: [2, 2], label: 'First'}, 
    {position: [2, 2], label: 'Second'},
    {position: [2, 2], label: 'Third'},
    {position: [2, 2], label: 'Fourth'},
    {position: [4, 1], label: 'Loner'}
  ]
}
```

### Preview
screenshot from 2018-08-13 09-42-19

### TODO
* [x] Docs
* [ ] Improve API (might be broken to issue so we have much smoother way of merging this things)


ryelo:
The red color here doesn't pass the WCAG AA color contrast test with normal text. To solve this, you could try bolding the text and increasing the font-size to `16px`

ryelo:
Also, the label on the x axis should go "Low | High" instead of "High | Low"

karelhala:
@ryelo really great catches! Currently if you copy + paste docs to any cmp it will look like this:
screenshot from 2018-08-20 15-57-28

Colors used here are: `['#f0ab00', '#FF6666', '#d1d1d1', '#bee1f4']` they should pass both `WCAG AAA` and `WCAG AA` as well.

PR #53, opened 2018-08-14
Enhancements/battery rewrite

ryelo:


src/PresentationalComponents/Battery/battery.js
```diff
@@ -6,30 +6,89 @@ import classNames from 'classnames';
...
+                </svg>
+            </i>
+        );
+    }
+
+    function batteryLevels (severity) {
+        switch (severity) {
+            case 'critical':
+            case 4:
+                return <path d="M16,212h416c8.8,0,16-7.2,16-16v-40c0-8.8-7.2-16-16-16H16c-8.8,0-16,7.2-16,16v40C0,204.8,7.2,212,16,212z M16,372h416
```
karelhala: Can you break these into their own components? Something like `CriticalBateryIcon`, `HighBateryIcon`, `WarnBateryIcon` and `InfoBateryIcon`.

This way we can do changes to just one icon and won't be bothered about the others.

ryelo:
@karelhala updated! The only question I have is if I need to `export * from './PresentationalComponents/Battery';` or if I can still use `export { default as Battery } from './PresentationalComponents/Battery/battery.js';` in the overall `index.js`

ryelo:
@karelhala this is hopefully the final commit that includes the correct exports as well as updated docs

PR #56, opened 2018-08-16
Inventory

karelhala:

PR #60, opened 2018-08-21
Add redux api for inventory and make it pluggable <?.x.?>

karelhala:
This PR is for pluggability of basic inventory, it has redux with API calls. Consumer will just import these redux functions and will be able to dispatch actions for fetching basic inventory information.


karelhala:
This PR is partial of https://github.com/RedHatInsights/insights-frontend-components/pull/56 so it's not that huge and easier to maintain.

src/redux/action-types.js
```diff
@@ -0,0 +1,23 @@
...
+  'LOAD_ENTITIES',
+  'LOAD_ENTITY'
+].reduce((acc, curr) => [
+  ... acc,
+  ...[curr, `${curr}_PENDING`, `${curr}_FULFILLED`, `${curr}_REJECTED`]
+], [])
+
+
+export const ACTION_TYPES = [
+    ...asyncActions
```
jharting: @karelhala the indention of the reduce step here was so that one can add additional strings into the array and have them turned into action types of the same name without having to specify both the name and value (which are same).

i.e. instead of
```js
export const SELECT_ENTITY = 'SELECT_ENTITY';
```

you can do

```js
export const ACTION_TYPES = [
    ...asyncActions,
    'SELECT_ENTITY',
    'CHANGE_SORT',
    ...
```

karelhala: Well with this approach you are kinda loosing some import goodies, it's really great for autogenerated items (async actions), but for static ones, you'd have to import `ACTION_TYPES` and then you'd loose check if `SELECT_ENTITY` is imported or not. If such action is changed or removed you will not be notified on build time, which is one of the main reasons for these constants.

jharting: @karelhala oh I see. I was not aware of the ES module export limitation

karelhala:
@jharting can your re-review this please? Not much of an updates, I just added new lines to end of each file.

PR #64, opened 2018-08-23
Features/gauge

Kinlaw:
Adds gauge component


Kinlaw:
@ryelo do you mind re-approving? It looks like your push reset the need for approval.

PR #65, opened 2018-08-28
Fix table issues with multiple sort icons and update styles

karelhala:


karelhala:
@ryelo I accidentaly created folder with same styles twice so I am removing such folder in this repo as well, that's why there is so much deletions.

ryelo:
Can we port over your examples to a storybook?

karelhala:
Sure thing, I'll update the example in storybook

PR #66, opened 2018-08-29
Use older version of node sass so we can use css vars assign

karelhala:
There are problems with newer version of node sass that when assigning sass variables into css4 variables they are not translated and such variables are broken. This eventually made tables look ugly because there are new variables related to table styling.

If anybody's interested about background of this issue https://github.com/sass/libsass/issues/2621


ryelo:
I agree that this is an issue, but IMO we should find another way to do a fix. It seems like that's the intended use case how it is now and is likely not to be changed in the future as the old way of doing it was a bug.

karelhala:
@ryelo sadly that's not on our side, but rather on pf-next :( I'll try to look around if I can find simpler solution

ryelo:
@karelhala do we want to merge this for now until we come across a fix/the fix is implemented?

karelhala:
I think we should go for the merge so consumers can use much better looking tables and once PF merges PR with all changes we can update the node sass again

PR #68, opened 2018-08-30
<?.?.x> remove gauge text and optimize id generator

Kinlaw:


src/Charts/Gauge/Gauge.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import propTypes from 'prop-types';
 import * as c3 from 'c3';
+import { select } from 'd3';
```
ryelo: Is this needed?

src/Charts/Gauge/Gauge.js
```diff
@@ -77,18 +82,15 @@ class Gauge extends Component {
 
 export default Gauge;
 
+/**
+ * generate random ID if one is not supplied
+ */
 /**
  * generate random ID if one is not supplied
  */
```
ryelo: Double comment

src/Charts/Gauge/Gauge.js
Kinlaw: Nope :(

Kinlaw:
@ryelo fixed

ryelo:
LGTM

src/Charts/Gauge/Gauge.js
```diff
@@ -36,6 +36,9 @@ class Gauge extends Component {
             gauge: {
                 fullCircle: true,
                 label: {
+                    format: function () {
```
karelhala: We don't need this function, do we?

Kinlaw: this removes the value in the center of the gauge whereas the `show: false` removes the min/max labels in the gauge. We want to remove the value in the center of the gauge in favor of creating our own data in the center so that we have more control of it.

PR #71, opened 2018-09-03
Add decorator for simpler use of registry

karelhala:


src/Utilities/Registry.js
```diff
@@ -0,0 +1,21 @@
...
+  return function(target) {
+    target.prototype.registry = init(settings.initialState, settings.middleware);
+  }
+}
+
+export function getStoreFromRegistry(initialState = {}, middleware = []) {
+  if(!registry) {
+    registry = new ReducerRegistry(initialState, [...middleware]);
+  }
+  return registry.getStore();
```
martinpovolny: Thanks, @karelhala, this simplifies the setup for Insights UI newcommers!

It feels ugly to me that the line 6-9 are the same as 17-20 except for the getStore() call. Should be no problem extraction is so that the currend `init` can be used in `getStoreFromRegistry`.

src/index.js
```diff
@@ -8,9 +8,10 @@ export {
     default as ReducerRegistry,
     applyReducerHash,
     dispatchActionsToStore,
-    reduxRegistery
+    reduxRegistery,
```
martinpovolny: @karelhala : I suggest replacing all "Registery" (notice the **E**) with "Registry". People will fight with this in their editors if it stays as it is ;-)

PR #73, opened 2018-09-04
Allow updating data for guage and donut charts <?.x.?>

karelhala:
* When data are changed for guage or donut chart automatically reload these charts.
* Mark data for charts as required.


ryelo:
Looks awesome!

PR #82, opened 2018-09-06
PF4 PageHeader and Main + Dark mode <?.x.?>

ryelo:


ryelo:
These are both layouts which are a little bit different, but I agree that there should be a way to do this without having to apply a dark prop to every component we have.

PR #83, opened 2018-09-09
Fix doc links in README <?.?.x> [skip ci]

martinpovolny:
ping @karelhala


karelhala:
Nice catch! Btw wrong style of commit, if you don't want to trigger release you should add `<no>` to the commit msg [commit-formatter](https://github.com/RedHatInsights/insights-frontend-components#used-formatter)

martinpovolny:
Fixed. Thx!

PR #86, opened 2018-09-10
Fix breadcrumbs example. <no>

martinpovolny:
Should be array of objects.


karelhala:
Thanks for the typo fix!

PR #87, opened 2018-09-10
Adding breakpoints on Donut Legend

ryelo:


src/Charts/Donut/donut.scss
```diff
@@ -44,15 +44,25 @@
...
 
 // with-legend
+
+// When larger than 0px, align the legend in a 1 column list
+@media only screen and (min-width: 0px) {
+    .ins-l-donut__legend { grid-template-columns: auto; }
+}
+
+// When larger than 992px, align the legend in a 2 column list
+@media only screen and (min-width: 992px) {
```
karelhala: Pf variable for large screen might be better here:
```scss
@media only screen and (min-width: var(--pf-global--breakpoint--lg)) {
  ...
}
```

karelhala:
Looks really great!

PR #88, opened 2018-09-12
Add a Chart pie

martinpovolny:
I need a pie chart. So starting a Pie component by copying the Donut one.

Ping @karelhala. Let's talk if this is a way forward or it's planned to happen differently.

### DIFF
--- a/doc/charts/pie.md
+++ b/doc/charts/pie.md
@@ -2,6 +2,20 @@
 
 This component is very similar to the [Donut chart](donut.md). Except there's no total and totalLabel.
 
+## A simple example
+
+```jsx
+
+let pieValues = [
+    ['value1', 5],
+    ['value2', 10],
+    ['value3', 20],
+    ['value4', 40]
+];
+
+<Pie values={pieValues} identifier='my_cool_pie' withLegend={true} />
+```
+
 ## Props
 
 ```JS
--- a/doc/charts/pie.md
+++ b/doc/charts/pie.md
@@ -1,0 +1,17 @@
+# Pie Chart
+
+This component is very similar to the [Donut chart](donut.md). Except there's no total and totalLabel.
+
+## Props
+
+```JS
+{
+    className: propTypes.string,
+    height: propTypes.number,
+    identifier: propTypes.string,
+    values: propTypes.array,
+    width: propTypes.number,
+    withLabel: propTypes.bool,
+    legendPosition: propTypes.oneOf(Object.keys(LegendPosition))
+}
+```
--- a/src/Charts/Pie/Pie.js
+++ b/src/Charts/Pie/Pie.js
@@ -118,14 +118,12 @@ class Pie extends Component {
         }
 
         return (
-            <React.Fragment>
-                <div className={wrapperClasses}>
-                    <div className='ins-l-pie'>
-                        <div id={this.props.identifier} className={pieClasses}></div>
-                    </div>
-                    {pieLegend}
+            <div className={wrapperClasses}>
+                <div className='ins-l-pie'>
+                    <div id={this.props.identifier} className={pieClasses}></div>
                 </div>
-            </React.Fragment>
+                {pieLegend}
+            </div>
         );
     }
 }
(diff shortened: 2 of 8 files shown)


src/Charts/Pie/Pie.js
```diff
@@ -0,0 +1,172 @@
...
+import propTypes from 'prop-types';
+import { generate } from 'c3';
+import { select } from 'd3';
+import classNames from 'classnames';
+import { Link } from 'react-router-dom';
+import isEqual from 'lodash/isEqual';
+
+import './pie.scss';
+
+export const LegendPosition = {
```
karelhala: No need to define such const here, you can import `LegendPosition` from Donut chart.

src/Charts/Pie/Pie.js
```diff
@@ -0,0 +1,172 @@
...
+            })
+        }
+    }
+
+    _updateChart () {
+        let data = {
+            type: 'pie',
+            columns: this.props.values,
+            pie: {
+                title: 'Pie'
```
karelhala: Title should come from props so consumers can change it.

martinpovolny:
> No need to commit demo files so others won't have merge conficts.

Sure. The demo is for my or reviewer's convenience. I'll remove it with the "WIP" label.

martinpovolny:
I removed the duplicate `LegendPosition` and remove the title from both Donut as Pie as it was AFAIK unused.

I'd leave the extraction of common code for a separate PR. Maybe as we will be adding more chart types.

src/Charts/Pie/Pie.js
```diff
@@ -0,0 +1,163 @@
...
+                                <span className="badge__number">({oneItem[1]})</span>
+                            </ConditionalLink>
+                        </div>
+                    </div>
+                ))}
+            </div>;
+        }
+
+        return (
+            <React.Fragment>
```
ryelo: No `React.Fragment` needed here

src/Charts/Pie/pie.scss
```diff
@@ -0,0 +1,105 @@
+@import '../../Utilities';
+@import '../../Utilities/variables';
```
ryelo: You shouldn't need to import `../../Utilities/variables` because `Utilities.scss` imports that for you.

src/Charts/Pie/pie.scss
```diff
@@ -0,0 +1,105 @@
...
+        &__number { @include rem(margin-left, 5px); }
+    }
+}
+
+.ins-l-pie-wrapper {
+    display: grid;
+    align-items: center;
+    justify-content: center;
+    grid-column-gap: 15px;
+    //grid-template: 'pie' 'legend';
```
ryelo: Can we remove this comment?

doc/charts/pie.md
```diff
@@ -0,0 +1,17 @@
...
+{
+    className: propTypes.string,
+    height: propTypes.number,
+    identifier: propTypes.string,
+    values: propTypes.array,
+    width: propTypes.number,
+    withLabel: propTypes.bool,
+    legendPosition: propTypes.oneOf(Object.keys(LegendPosition))
+}
+```
```
ryelo: Can we add some examples here, even though they're very similar to the Donut?

src/Charts/Pie/Pie.js
martinpovolny: fixed in both Donut and Pie

src/Charts/Pie/pie.scss
martinpovolny: fixed in both Donut and Pie

src/Charts/Pie/Pie.js
ryelo: Donut still needs it because it has two sibling tags

src/Charts/Pie/pie.scss
martinpovolny: fixed in both Donut and Pie

doc/charts/pie.md
martinpovolny: Sure.

martinpovolny:
@ryelo : is one example enough or should I mimic what is in Donut?

ryelo:
@martinpovolny One example should be fine!

martinpovolny:
missed that :-(

here it is: https://github.com/RedHatInsights/insights-frontend-components/pull/91

thx!

karelhala:
It's fine, thanks for the PR. Merging

PR #90, opened 2018-09-17
Update dropdown to be used from react-core <?.x.?>

karelhala:
* Update all packages to newer version
* Use dropdown from react-core


ryelo:
@karelhala can you fix the broken Travis and I'll review?

karelhala:
Oh darn it, I forgot to add dependency on font awesome sass. Fixed that

ryelo:
@karelhala looks like some variables are missing now

karelhala:
@ryelo these problems are caused by some changes in pf-react where they removed these dependencies. I reinstalled them and added them into `package.json` so we are sure that they are installed in consumes app as well.

PR #94, opened 2018-09-20
Data driven form

Hyperkid123:
There was a discussion about data-driven form component, that should be used in new projects. @lgalis and @mkanoor had contacted me and after some time we have decided to implement this component.

There is an incentive to use [mozilla json schema](https://github.com/mozilla-services/react-jsonschema-form) for this because some other teams have been using it(Openshift) and the swagger-api is supporting this out of the box. After some discussion we have decided to do the following.

### JSON-SCHEMA
We will use the json format, but that is it. The library itself is usable for most simple use-cases (Form with x amount of field). It supports some basic dynamic features, but for most of advanced forms (tabs, wizards, conditional fields etc.) you have to write some custom code that mutates the whole library. Also the form components are not usable, because they are not PF4. You can customize the look and feel of the form components, but at that point, you are using only the JSON format and the form API.

It's not possible to use the mozilla formApi for custom forms.

### Custom rendering
Because most of the library would not be used and we need some way to customize our forms further, it was decided that custom rendering component will be used. That component must support the mozilla JSON-SHCEMA (to support other teams and swagger) and we will extend it for our needs and with our (PF4) form components.

### Work done so far
We are starting with use cases from the Service Catalog team. Component will be used by both Service Catalog and Sources teams. As we progress with adding more providers we will implement the required additional functions.

### Goals
- data-driven forms
- styleable
- PF4 compatible
- swagger compatible
- support mozilla json schema
- support MIQ service catalog as well as provider forms
- support extend ability by provider teams as implemented in MIQ 

cc @karelhala @martinpovolny


src/App.scss
```diff
@@ -0,0 +1,2 @@
+$pf-global--enable-font-overpass-cdn: true;
```
karelhala: No need to have this file, check the latest master this file is strill not relevant.

src/forms/DataDrivenForm/componentMapper.js
```diff
@@ -0,0 +1,20 @@
...
+  password: { ...component, type: 'password' },
+})[type];
+
+export default (type, widget) => {
+  let component = typeMapper(type);
+  if (widget !== undefined) {
+    component = widgetMapper(widget, component);
+  }
+  return component;
+};
```
karelhala: Missing EOL

src/forms/DataDrivenForm/componentMapper.js
```diff
@@ -0,0 +1,20 @@
+import { TextField, TextareaField, SwitchField } from '../FormFields/formFields';
+
+const typeMapper = (type = 'string') => ({
+  string: { component: TextField, type: 'text' },
+  integer: { component: TextField, type: 'number' },
+  boolean: { component: SwitchField },
+})[type];
+
+const widgetMapper = (type, component) => ({
```
karelhala: I love this typeMapper! Can the widget mapper and type mapper be merged together?

src/forms/DataDrivenForm/formRenderer.js
```diff
@@ -0,0 +1,106 @@
...
+        autoFocus={options.autoFocus === key}
+        isRequired={globalValidators[key] && globalValidators[key].name === 'required'}
+      />
+    </GridItem>
+  );
+});
+
+const renderTitle = (title, description) => (
+  <div>
+    {title && <Fragment><Title size="3xl">{title}</Title><hr /></Fragment>}
```
karelhala: Can we omit the `<hr />` tag here? I guess this will be styled by consumers and it will be quite painfull to remove it from DOM by css.

src/forms/DataDrivenForm/formRenderer.js
```diff
@@ -0,0 +1,106 @@
...
+    {description && <Title size="lg">{description}</Title>}
+  </div>
+);
+
+const setGlobalValidation = (validators) => {
+  const result = {};
```
karelhala: Can this be simplified by map? For example like this:
```JS
const setGlobalValidation = (validators) => {
  return validators.map(() => validationMapper('required'))
}
```

src/forms/DataDrivenForm/formRenderer.js
```diff
@@ -0,0 +1,106 @@
...
+  validators.forEach((item) => {
+    result[item] = validationMapper('required');
+  });
+  return result;
+};
+
+const createInitialValues = properties => {
+  const initialValues = {};
```
karelhala: Could the `forEach` here be simplified by `map`?

src/forms/DataDrivenForm/formRenderer.js
```diff
@@ -0,0 +1,106 @@
...
+            </form>
+          );
+        }}
+      />
+    </Fragment>
+  );
+  }
+}
+FormRenderer.propTypes = {
+schema: PropTypes.object.isRequired,
```
karelhala: Missing space here.

src/forms/DataDrivenForm/formRenderer.js
```diff
@@ -0,0 +1,106 @@
...
+uiSchema: PropTypes.object,
+onSubmit: PropTypes.func.isRequired,
+};
+
+FormRenderer.defaultProps = {
+uiSchema: {},
+};
+
+export default FormRenderer;
```
karelhala: Missing EOL.

src/forms/DataDrivenForm/componentMapper.js
```diff
@@ -0,0 +1,20 @@
+import { TextField, TextareaField, SwitchField } from '../FormFields/formFields';
```
karelhala: `TextareaField` is unused

src/forms/FormFields/formFields.js
```diff
@@ -0,0 +1,50 @@
...
+  helperText: PropTypes.string,
+  meta: PropTypes.object,
+}
+
+TextField.defaultProps = {
+  type: 'text',
+  isRequired: false,
+  helperText: undefined,
+  meta: {},
+}
```
karelhala: Missing EOL

src/forms/Helpers/helpers.js
```diff
@@ -0,0 +1,6 @@
+export const composeValidators = (...validators) => value =>
+  validators.reduce((error, validator) => error
+    || (typeof validator === 'function'
+      ? validator(value)
+      : undefined),
+    undefined);
```
karelhala: Missing EOL.

src/forms/Helpers/helpers.js
```diff
@@ -0,0 +1,6 @@
+export const composeValidators = (...validators) => value =>
+  validators.reduce((error, validator) => error
```
karelhala: You are basically doing some in here, am I right? Since you are always sending the first `error` you encounter it would be better to use [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) so when we encounter first error we will show it to user.

src/forms/Validators/validators.js
```diff
@@ -0,0 +1,22 @@
...
+  return value => value && value.length > 0 ? undefined : message
+});
+
+export const minLength = treshold =>
+  value =>
+    !value || value.length === 0
+      ? undefined
+      : value.length >= treshold
+        ? undefined
+        : `Should be atleast ${treshold} long`;
```
karelhala: Missing EOL

src/forms/Helpers/helpers.js
Hyperkid123: it would be nice but i actually need the error message and some returns only `true/false`

Hyperkid123:
@karelhala is there anything else?

PR #96, opened 2018-09-20
Adding Linting

ryelo:
**Please review thoroughly** - I looked at it and everything seemed fine, but I just need more eyes to double check.

I saw that we were having inconsistencies in our `.js` files where the indenting was off as well as other things. This will standardize this component library with the rest of the insights applications to have cleaner code.

Another PR should be opened up in the future regarding file naming.

@karelhala can you pay close attention to: [Table Footer](https://github.com/RedHatInsights/insights-frontend-components/blob/1b3b3053065dfccea0d7c3bb6014a02b1b35c0ef/src/PresentationalComponents/Table/Table.js#L79) - this was causing lint to throw errors about passing a children prop inside the tag instead of using the children between tags. I don't really have a preference, but we can turn that rule off if needed.

There is also a `console.log` in the pagination that I left, but feel free to take it out as it only throws a warning.

cc: @martinpovolny


.babelrc
```diff
@@ -1,7 +1,16 @@
 {
-    "presets": ["./.babelrc.js"],
```
karelhala: Is there a reason why we can't use presets from local JS file? Because that will break linking of this package.

ryelo: Nope. I'll add it back!

ryelo:
@karelhala should be good now. Travis had some caching issues and it was failing builds 😞

.babelrc
```diff
@@ -1,7 +1,16 @@
...
+    "presets": ["env", "react", "es2017", "es2015", "./.babelrc.js"],
     "ignore": [
         "./index.js"
+    ],
+    "plugins": [
```
karelhala: Well we don't really need the plugins and presets here since these are covered by `.babelrc.js`.

ryelo:
@Kinlaw do you mind reviewing this too? Just wanting to make sure I don't break all the components

ryelo:
@karelhala mind reviewing again? Had to rebase.

Issue #99, opened 2018-09-21
File Naming


ryelo:
@karelhala @Kinlaw 

How do we want to handle naming of files?

I was thinking:
```
PresentationalComponents
└───YourComponent
    ├─── index.js
    ├─── YourComponent.js
    └─── styles.scss <- literally named styles.scss
```

karelhala:
That or
```
PresentationalComponents
└───YourComponent
    ├─── index.js
    ├─── YourComponent.js
    └─── YourComponent.scss
```

I guess both are acceptable and could be used at the same time

PR #101, opened 2018-09-24
renaming files to standardize

ryelo:
#99


ryelo:
👎 to MacOS case sensitivity

Will squash when merging.

PR #102, opened 2018-09-24
Switch Dark theme to use ThemeContext <?.x.?>

ryelo:
instead of specifying a `dark` prop or something on each component, this will allow you to wrap the page in a `<Dark>` component tag which will apply the appropriate classes to children components using React's `Context`.

The default context is going to be `light` and the flip of that is obviously `dark`. However, because some components don't use `pf-m-light` or a different version of `pf-m-dark-foo`, they have to be specified.


src/PresentationalComponents/Dark/Dark.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import propTypes from 'prop-types';
+
+import ThemeContext from './configContext';
+
+class Dark extends React.Component {
```
karelhala: Optional: No need to use class here
```JSX
import React from 'react';

import ThemeContext from './configContext';

const Dark = ({ children, ...props }) => (
  <ThemeContext.Provider {...props} value='dark'>
    {children}
  </ThemeContext.Provider>
)
```

src/PresentationalComponents/Dark/configContext.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+
+const ThemeContext = React.createContext('light');
```
karelhala: Optional: You can make this shorter by exporting just the context
```JSX
import React from 'react';

export default React.createContext('light');
```

src/PresentationalComponents/PageHeader/PageHeader.js
```diff
@@ -0,0 +1,43 @@
...
+const PageHeader = ({ className, children, ...props }) => {
+
+    let pageHeaderClasses = classNames(
+        className,
+        'pf-l-page__main-section'
+    );
+
+    return (
+        <ThemeContext.Consumer>
+            { darkTheme => {
```
karelhala: Rename this to `theme` to be more correct
```JSX
<ThemeContext.Consumer>
  { theme => {
      let themeClasses = classNames(
        { [`pf-m-${ theme }-200`]: theme  === 'dark' },
        { [`pf-m-${ theme }`]: theme  === 'light' }
      );
      //..
    }
  }
</ThemeContext.Consumer>
```

Issue #108, opened 2018-09-26
Add widget types and IDs to Components


ryelo:
### Charts
- [x] Donut
- [x] Gauge
- [x] Matrix
- [x] Pie

### Components
- [x] Ansible
  - won't need unique ID because it should be wrapped in a component with an ID already.
- [x] Battery (needs unique ID)
- [x] Breadcrumbs
- [x] Dropdown (needs unique ID)
- [x] Input (needs unique ID)
- [x] PageHeader
- [x] Pagination
- [ ] SimpleTableFilter (needs unique ID)
- [x] TabLayout
- [x] Table (needs unique ID)

PR #109, opened 2018-09-26
Adding widget attributes <?.x.?>

ryelo:
#108


src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
```diff
@@ -11,7 +11,7 @@ function upperCaseFirst(item)
 const Breadcrumbs = ({ items, current, className, onNavigate, ...props }) => (
     <React.Fragment>
         {
-            items.length > 0 && <ol { ...props } className={ classnames(className, 'ins-breadcrumbs') }>
+            items.length > 0 && <ol { ...props } className={ classnames(className, 'ins-breadcrumbs') } widget-type='InsightsBreadcrumbs'>
                 { items.map((oneLink, key) => (
                     <li key={ oneLink.navigate }>
```
karelhala: Can you add here `data-key={key}` as well? So Qe can query much easily

PR #123, opened 2018-10-04
Push lock file as well when auto releasing <no> [skip ci]

karelhala:
This was brought by @AllenBW when we were discussing updating lock file dependencies. Not sure if we should push it tho, so I wanted to bring this topic for discussion.
#### Pros
 - No need to push new lock file when running `npm install`

#### Cons
 - We have to run `npm install` multiple times (`npm ci` does not updates package-lock file)
 - Automatically updated lock file might break our builds


AllenBW:
Would like to add another pro,
* Ensures the `package.json` and `package-lock` reference the same version (which might not matter much cuz the primary consumers of ifc is insights and redhat friends... but just imagine... if pfr or lodash or any other repo had that mismatch 😱 )

ryelo:
I think this was a discussion we had at the beginning of the project, but I would love to stick with `npm ci` for the travis environment.

It's faster, it's removes the node_modules folder before installing, and it ensures that `package` and `package-lock` are running the same things.

AllenBW:
bahahaha first sentence "Make sure you have a package-lock and an up-to-date install:" reads like a winner to me 😋

ryelo:
Any update on this or can we close?

AllenBW:
Can we merge it? 🤔

karelhala:
Sorry for the late response, but I was kinda against pushing lock file on every release, but looks like everytime we release new version and someone does `npm install` it will bump the version and we have inconsistent lock version and package version.

PR #131, opened 2018-10-15
Add support for another select definition and conditional fields.

Hyperkid123:
Added support for reference field definition and alternative select definition for the FormRenderer. This change is base on this PR: https://github.com/ManageIQ/insights-api-service_portal/pull/15


Hyperkid123:
cc @karelhala

PR #134, opened 2018-10-17
Reverts pr #133 

AllenBW:
Cuz we really really hate developers that give the battery component invalid input


AllenBW:
#135 makes this unnecessary

PR #135, opened 2018-10-17
adding null battery

ryelo:
@karelhala @AllenBW 

This should render a null battery if there is no data. Will this work?


AllenBW:
OMG @ryelo YER DA BEST <3

PR #145, opened 2018-11-13
Fixed notifications styling and error response.

Hyperkid123:
Fixes notification width for smaller screens and made it a bit bigger on desktop size screens.

Also added the possibility, that error response is string, not object.


Hyperkid123:
cc @karelhala

PR #146, opened 2018-11-13
Use payload from action in notifications

karelhala:


karelhala:
Yes, we want. Since we want to have zero config out of the box notifications if some error happens this will help up greatly.

ryelo:
Ah okay. Sounds good 🎉

PR #153, opened 2018-11-26
Fixing prop spread on Wizard <?.?.x>

ryelo:


src/PresentationalComponents/Wizard/Wizard.js
```diff
@@ -50,7 +50,6 @@ class Wizard extends Component {
 
         return (
             <Modal
-                { ...this.props }
```
karelhala: Do not use spread in this way please, it will try to add props which are not recognized by react
```
// Notice the handleModalToggle being taken away from rest of props
const { isLarge, title, className, isOpen, handleModalToggle, ...props } = this.props;
// ...
<Modal
  {...props}
  isLarge = { isLarge }
  title= { title }
  className= { className }
  isOpen={ isOpen }
  onClose={ this.handleOnClose }
  actions={ renderModalActions }>
    { this.props.content[this.state.currentStep] }
</Modal>
```

PR #154, opened 2018-11-28
Adding babel plugins to add displayName to components <?.?.x>

ryelo:


package.json
```diff
@@ -55,13 +55,15 @@
     "fontAwesome": "node_modules/font-awesome-sass/assets/stylesheets"
   },
   "devDependencies": {
+    "@babel/plugin-transform-react-display-name": "^7.0.0",
```
karelhala: Use older version of plugin to make it works with older version of babel https://www.npmjs.com/package/babel-plugin-transform-react-display-name

PR #158, opened 2018-12-04
Fix SimpleTableFilter use of Dropdown.

martinpovolny:
SimpleTableFilter: use the new way of passing DropdownItems to the Dropdown component.

http://patternfly-react.surge.sh/patternfly-4/components/dropdown

Fixes broken layout of the drop down list.


martinpovolny:
ping @karelhala

PR #159, opened 2018-12-04
Add path prefix and base path option to inventory

karelhala:
Allow passing `pathPrefix` to indicate version of inventory and `apiBase` to change URL from which data are fetched.


karelhala:
@dLobatog feel free to review as well

PR #160, opened 2018-12-05
Fix Table expand button missing arial-label.

martinpovolny:
Add a missing `arial-label` to row expand button to fix a warning in the browser console:

```
checkPropTypes.js:19 Warning: Failed prop type: aria-label is required for Buttons with the plain variant
    in Button (created by TableBody)
    in TableBody (created by Table)
    in table (created by Table)
```

Ping @karelhala

### DIFF
--- a/src/PresentationalComponents/Table/TableBody.js
+++ b/src/PresentationalComponents/Table/TableBody.js
@@ -43,6 +43,7 @@ class TableBody extends Component {
                         'pf-m-expanded': row.active
                     }) }
                     onClick={ event => this.props.onExpandClick && this.props.onExpandClick(event, row, key) }
+                    aria-label={ `Expand row ${key}` }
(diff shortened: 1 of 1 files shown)


src/PresentationalComponents/Table/TableBody.js
```diff
@@ -43,6 +43,7 @@ class TableBody extends Component {
                         'pf-m-expanded': row.active
                     }) }
                     onClick={ event => this.props.onExpandClick && this.props.onExpandClick(event, row, key) }
+                    aria-label='expand a row'
```
karelhala: Indication which row is going to be expanded might be better
```JSX
aria-label={`Expand row ${key}`}
```

martinpovolny:
@karelhala : updated

PR #164, opened 2018-12-07
Change inventory detail to be as special component

karelhala:
This will change clicking on inventory table row to navigate to special inventory detail.


karelhala:
ping @jiridostal @AllenBW @dLobatog

PR #165, opened 2018-12-10
Added high z-index to notifications container.

Hyperkid123:
@martinpovolny had an issue, that the notifications were rendered under the insights APP toolbar. I've added z-index value to the notifications container.


Hyperkid123:
cc @karelhala

src/SmartComponents/Notifications/notifications.scss
```diff
@@ -2,6 +2,7 @@
   position: fixed;
   top: 0;
   right: 0;
+  z-index: 100;
```
karelhala: Too low number, main content has z-index set to `100` so notifications would be still hidden.

ryelo:
@Hyperkid123 make sure to format so that the semantic release picks this up too.

Issue #149, opened 2018-11-15
Pagination component page input needs debounce


AllenBW:
Title says it all, on input, instantly whatever function is tied to the input, is executed, there should be a little pause on input so users can type long numbers 😋 

![test](https://user-images.githubusercontent.com/6640236/48562822-fc5b3200-e8c0-11e8-8511-b4aadba1e840.gif)

karelhala:
Sadly this cannot be fixed in frontend components because of how pagination works with input. If we'd were to add debounce here page number wouldn't change until debounce finishes, that is a no-go and is really bad. We could work around this, but it would add a lot of custom logic that would need to be updated once we start using PF4 pagination. However I am gonna write doc example on how to work with this because now we are passing additional argument to tell consumers that they should debounce.

karelhala:
https://github.com/RedHatInsights/insights-frontend-components/pull/248

AllenBW:
Drat, I was hoping this wouldn't be the solution, means more changes in other places, but thanks for following up @karelhala !! #248 is a great piece of work 🤗

PR #168, opened 2018-12-12
[WIP] Filter and pagination inside inventory

karelhala:
This PR adds two crucial things to inventory
 * filters with default values
 * pagination

There is also new callback for inventory table that listens on refreshing data - refresh, filter change, pagination change


karelhala:
ping @dlobatog @jdostal @AllenBW

ryelo:
@karelhala When there are >9 pages, will this issue still occur? https://github.com/RedHatInsights/insights-frontend-components/issues/149

karelhala:
@ryelo sadly, yes :( But I am gonna work on PR that fixes such issue by using debounce from lodash.

src/api/inventory.js
```diff
@@ -63,8 +63,40 @@ const mapData = ({ results, ...data }) => ({
...
+        ].filter(Boolean);
+    }
+
+    return query ? `?${query.join('&')}` : '';
+}
+
+export function getEntities(items, { base = INVENTORY_API_BASE, ...rest }) {
+    let query = buildQuery(rest);
+
+    return fetch(`${base}${items.length !== 0 ? '/' + items : ''}${query}`).then(r => {
```
dLobatog: I believe this is building the wrong query for 'tree-like' Inventory tables. Notice in such tables, you would have an items array like the following: https://github.com/RedHatInsights/insights-frontend-components/pull/168/files#diff-99c7484e5cb238b566e93f5654bae0e7R290

All the collapsible rows don't have any ID, therefore, this is constructing a query with 'some-id,' (finishing with a comma), which returns a 500 from the inventory.

A simple `items.filter(n => n)` here fixes the issue.

src/SmartComponents/Inventory/EntityTable.js
```diff
@@ -87,20 +112,46 @@ class EntityTable extends React.Component {
...
+            ...cells,
+            health,
+            actions
+        ].filter(Boolean);
+    }
+
     render() {
-        const { columns, entities, rows, showHealth, loaded } = this.props;
+        const { columns, entities, rows, showHealth, loaded, expandable, onExpandClick } = this.props;
         const filteredData = (entities || rows).filter(oneRow => oneRow.account);
```
dLobatog: I believe in this case, the OR statement does no longer apply. It's either entities (if we're using a regular, non tree-like table), or rows (if we're using a expandable table). `entities` contains the data from the inventory, so it won't contain any additional collapsible rows. This means that even if we provide the `rows` with extra info, they will get discarded in favor of entities, rendering the 'expandable' attribute useless. See:

screenshot from 2018-12-28 11-00-36

doc/components/inventory.md
```diff
@@ -262,6 +271,163 @@ class SomeCmp extends React.Component {
...
+        const { items } = this.state;
+        items.find(item => item.id === key).active = !row.active;
+        // Not ideal, but for the sake of example it's fine
+        row.children.forEach(child => {
+            items.find(item => item.id === child.id).isOpen = !row.active;
+        });
+        this.setState({
+            items
+        });
+        this.context.store.dispatch(this.updateEntities(items));
```
dLobatog: `this.context.store` seems to be undefined?

src/api/inventory.js
dLobatog: Well it does fix the issue.. for the inventory table. For inventory detail, `items.filter` wouldn't work because `items` here is just an ID. I suppose it would have to be dealt with differently. Maybe making an array `[items]` then flattening it, so it works for both.

dLobatog:
```javascript
function flatten(a) {
    return Array.isArray(a) ? [].concat(...a.map(flatten)) : a;
}

export function getEntities(items, { base = INVENTORY_API_BASE, ...rest }) {
    let query = buildQuery(rest);

    return fetch(`${base}${items.length !== 0 ? '/' + flatten([ items ].filter(n => n)) : ''}${query}`)
 ```

this worked

File: src/SmartComponents/Inventory/EntityTable.js (hunk 2 above)
Comment (dLobatog) [reply src/SmartComponents/Inventory/EntityTable.js]: Hmm.. on the other hand, entities contains valuable info, such as 'facts'. Maybe the OR could be reversed and as a user I could merge 'entities facts' in 'rows' on the reducer.

File: src/api/inventory.js (hunk 1 above)
Comment (karelhala) [reply src/api/inventory.js]: This function expects items to be either one ID or array of IDs, there is function in [InventoryList.js#L97](https://github.com/RedHatInsights/insights-frontend-components/blob/f4c5b0b79195445652891d24772ff4bfd57897de/src/SmartComponents/Inventory/InventoryList.js#L97) which does something similiar to what you are proposing. So no need to do this twice since I wanted to keep this function as simple as possible.

File: src/SmartComponents/Inventory/InventoryList.js [hunk 4]
```diff
@@ -1,68 +1,66 @@
...
+const propTypes = {
     filterEntities: PropTypes.func,
     loadEntities: PropTypes.func,
     pathPrefix: PropTypes.number,
     apiBase: PropTypes.string,
     showHealth: PropTypes.bool,
+    page: PropTypes.number,
+    perPage: PropTypes.number,
+    onRefresh: PropTypes.func,
     items: PropTypes.oneOfType([
```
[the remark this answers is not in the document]
karelhala: The ID is optional, so no need to get rid of it. But adding another shape with coresponding data structure and marking this ID in current shape as required is good way!

src/SmartComponents/Inventory/EntityTable.js
karelhala: Those entities and rows are something which needs to be refactored, because both holds same information - data from server. Entites were ment to be filtered, sorted paginated data from server. But since we are doing pagination, filter and sort on server that is no longer applicable and items should be removed from store.

doc/components/inventory.md
karelhala: Context is automatically bound to this by calling constructor with second parameter and then passing it to parent using super https://github.com/facebook/react/blob/master/packages/react/src/ReactBaseClasses.js#L21

karelhala:
@dLobatog thanks for the review! I'll update the code based on your suggestions.

src/api/inventory.js
dLobatog: I don't think this works with one ID, really. It does work if it contains an array of IDs with one ID. Using just an ID, the "," at the end of the URL breaks the inventory API call. 

 GET https://prod.foo.redhat.com:1337/r/insights/platform/inventory/api/v1/hosts/ed9b9afd-4490-4ba8-b1a5-af63fd34c124, - 500 (Internal Server Error) (note the dangling comma)

screenshot from 2019-01-02 11-40-36

If you don't want to change this function, perhaps whatever calls `getEntities` should be aware that 'items' can look like [id1], or [id1, id2, id3], but not simply 'id1' as the inventory does not respond to that

karelhala: Ooh, I see your point now. Yeah items has to be array of strings, not just plain string, I am gonna add error throw if incorrect data type is sent to inventory to notify consumers how it should look like. Good catch however!

src/SmartComponents/Inventory/EntityTable.js
dLobatog: Do you plan to do that refactor here? I am wondering how to show expandable rows otherwise - so far I locally made the change `(rows || entities)` and I'm using my locally modified library.

doc/components/inventory.md
dLobatog: I know, `this.context` is bound properly, but `this.context.store` is undefined, where is it initialized?

PR #176, opened 2018-12-19
Added multiple notifications message key options

Hyperkid123:
Added option to specify multiple formats for error responses for automatic notifications.

Also i had to increase z-index of notification container once again. I will create probably a section in docs that will document z-index values in insights applications. So far i have these values:

- main-content: 100
- side-bar: 200
- page-header: 300
- account button in top right corner - the border had 1000 now it seems to be 300 also.


Hyperkid123:
cc @karelhala @ryelo

PR #179, opened 2018-12-20
Add pf-c-page to page elements, like main and header

karelhala:
With recent changes to pf-next page has moved from layout to component, we need to be aligned with both of these versions so when chrome picks up new version it's not going to blow up with badly styled elements.


ryelo:
LGTM, can you update the branch and then make sure that the Travis error is fixed?

karelhala:
@ryelo please retest and re-review.

PR #181, opened 2019-01-04
Add General info and get data from correct places

karelhala:
selection_005

### DIFF
--- a/src/SmartComponents/Inventory/applications/GeneralInformation.js
+++ b/src/SmartComponents/Inventory/applications/GeneralInformation.js
@@ -1,12 +1,22 @@
-import React, { Fragment } from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { TextContent, Text, TextVariants, Grid, GridItem, GutterSize } from '@patternfly/react-core';
+import {
+    Card,
+    CardHeader,
+    CardBody,
+    TextContent,
+    Text,
+    TextVariants,
+    Grid,
+    GridItem,
+    GutterSize
+} from '@patternfly/react-core';
 import { Table } from '../../../PresentationalComponents/Table';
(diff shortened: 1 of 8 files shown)


karelhala:
ping @ryelo

src/SmartComponents/Inventory/applications/general-information.scss
```diff
@@ -0,0 +1,18 @@
...
+        > div:first-child {
+            border-bottom: solid 0.1em var(--pf-global--primary-color--100)
+        }
+
+        div:not(:first-child) {
+            span:first-child {
+                font-weight: var(--pf-global--FontWeight--bold);
+            }
+            span:not(:first-child) {
+                float: right;
```
ryelo: No floats!!!
Let's use a flex property to accomplish the same thing.

src/SmartComponents/Inventory/applications/GeneralInformation.js
```diff
@@ -1,16 +1,117 @@
...
-            <div>General Information</div>
+            <Grid sm={ 6 } md={ 6 } lg={ 6 } gutter={ GutterSize.md }>
+                <GridItem>
+                    <TextContent>
+                        <Card>
+                            <CardHeader>
+                                <Text component={TextVariants.h1}>System</Text>
+                            </CardHeader>
+                            <CardBody>
+                                <div>
```
ryelo: Instead of using all these `<div>s` can we just use PF's `<Level>/<LevelItem>` component?

src/SmartComponents/Inventory/applications/general-information.scss
karelhala: If I'm gonna use Level there's not gonna be any need for floats.

ryelo:
@karelhala did you put in the `<Level>` component? I see you imported it and then referenced it in the .scss file, but I don't see it actually in the code.

karelhala:
@ryelo ah damn, yeah forgot to use it...

PR #186, opened 2019-01-09
Do not dispatch new action from router params if on same page

karelhala:
One nasty bug appears in advisor that the page gets into infinite loop of refreshes because multiple components change router params to same, but they treat it as different. This fixes such issue by comparing params and path and if either params or path is different do path update, if not do not fire such update.

### DIFF
--- a/src/api/inventory.js
+++ b/src/api/inventory.js
@@ -1,57 +1,5 @@
 export const INVENTORY_API_BASE = '/r/insights/platform/inventory/api/v1/hosts';
 
-/* eslint camelcase: off */
-function buildMock(i) {
-    return {
-        id: `${i}`,
-        canonical_facts: {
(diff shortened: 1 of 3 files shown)


src/api/inventory.js
```diff
@@ -96,16 +96,18 @@ function buildQuery({ per_page, page, filters }) {
...
+        () => fetch(`${base}${items.length !== 0 ? '/' + items : ''}${query}`).then(r => {
+            if (r.ok) {
+                return r.json().then(mapData);
+            }
 
-        // TODO: remove me
-        if (r.status === 404) {
-            return mapData(mockData(items));
-        }
+            // TODO: remove me
```
AllenBW: can we kill this cuz mock data is now gone?

karelhala: Sure thing!

PR #187, opened 2019-01-10
Compliance info

karelhala:
Add compliance information component to be shared amongst other applications.

#### Bug fixes:
 * row collapse possible on second and other pages
 * correct number of rules now available (not counting long description)
 * selecting on second and other pages corrected

#### New Features:
 * added cache invalidation when switching from different details (i.e. when component was destroyed)
 * loading is now made with content loader to have somewhat smoother loading indication

#### UI
selection_010
selection_011


karelhala:
@dLobatog yes, you'll use it as you are using your component so far. No need to change anything for you, just import this component instead of which you are using. If you are asking about the tabs in there, those are only for inventory app.

src/SmartComponents/Inventory/applications/compliance/compliance.scss
```diff
@@ -0,0 +1,8 @@
+.pf-c-card.ins-error-card div {
+    text-align: center;
+}
+
+#remediation-button > button {
+    padding-left: 10px;
+    padding-right: 10px;
```
ryelo: Can we use the rem mixin for this?

karelhala: Sure thing!

PR #189, opened 2019-01-10
Advisor info

karelhala:
selection_012

### DIFF
--- a/src/SmartComponents/Inventory/applications/advisor/Advisor.js
+++ b/src/SmartComponents/Inventory/applications/advisor/Advisor.js
@@ -96,8 +96,7 @@ class InventoryRuleList extends Component {
                             <FrownOpenIcon size='lg' />
                         </CardHeader>
                         <CardBody>
-                            There was an error while trying to fetch rules list for this entity. Please contact your
-                            administrator and show him this screen.
+                            There was an error fetching rules list for this Entity. Please show your administrator this screen.
                         </CardBody>
                     </Card>
                 ) }
--- a/src/SmartComponents/Inventory/applications/advisor/Advisor.js
+++ b/src/SmartComponents/Inventory/applications/advisor/Advisor.js
@@ -16,7 +16,8 @@ class InventoryRuleList extends Component {
     state = {
         expanded: true,
         inventoryReportFetchStatus: 'pending',
-        inventoryReport: {}
+        inventoryReport: {},
+        kbaDetails: []
     }
 
     componentDidMount() {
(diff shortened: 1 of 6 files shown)


AllenBW:
essential https://github.com/RedHatInsights/insights-advisor-frontend/pull/100 cept here, upon merging this, we're gonna have to revert https://github.com/RedHatInsights/insights-advisor-frontend/pull/100 in advisor frontend

src/SmartComponents/Inventory/applications/advisor/ExpandableRulesCard.js
```diff
@@ -0,0 +1,122 @@
...
+                        <SplitItem><Battery label='Risk Of Change' severity={ rule.resolution_risk } /></SplitItem>
+                    </Split>
+                </CardHeader>
+                { expanded && (<CardBody>
+                    <Grid gutter='md' sm={ 12 }>
+                        <GridItem>
+                            <Card className='pf-t-light  pf-m-opaque-100'>
+                                <CardHeader> <ThumbsUpIcon /> Detected Issues</CardHeader>
+                                <CardBody>
+                                    <div dangerouslySetInnerHTML={ { __html: rule.reason_html } } />
```
dkuc: FYI, we plan to remove these _html properties from the rule.  in this case you would just use .reason

src/SmartComponents/Inventory/applications/advisor/Advisor.js
```diff
@@ -0,0 +1,126 @@
...
+
+    expandAll(expanded) {
+        this.setState({ expanded: !expanded });
+    }
+
+    buildRuleCards = () => {
+        const { inventoryReport, expanded } = this.state;
+        return (
+            <Fragment>
+                <div className="pf-u-display-flex pf-u-flex-direction-row-reverse">
```
AllenBW: This doesn't appear to be showing up correctly, should be in upper right of the list

src/SmartComponents/Inventory/applications/advisor/Advisor.js
```diff
@@ -0,0 +1,126 @@
...
+                        </Card>
+                ) }
+                { inventoryReportFetchStatus === 'failed' && (
+                    <Card className="ins-empty-rule-cards">
+                        <CardHeader>
+                            <FrownOpenIcon size='lg' />
+                        </CardHeader>
+                        <CardBody>
+                            There was an error while trying to fetch rules list for this entity. Please contact your
+                            administrator and show him this screen.
```
AllenBW: Gender bias alarms going off, unless this is a uxd sanctioned message would recommend considering the following rewording: 

" There was an error fetching rules list for this Entity. Please show your administrator this screen."

src/SmartComponents/Inventory/applications/advisor/Advisor.js
karelhala: You ought to `import '@red-hat-insights/insights-frontend-components/components/Advisor.css'` as well.

karelhala: Oh I see, mmhh the problem here is that we are currently using old version of pf-next in FE components. So you have to import '@patternfly/patternfly-next/utilities/Flex/flex.css` on your own so this text is on right

AllenBW: 👍 🤗

karelhala:
@AllenBW updated, please retest and re-review.

PR #191, opened 2019-01-10
Detail styling

karelhala:
selection_013


src/SmartComponents/Inventory/EntityDetail.js
```diff
@@ -25,9 +36,90 @@ class EntityDetails extends Component {
...
+            </Grid>
+        );
+    }
+
+    generateFacts = () => {
+        return (
+            <Grid className="ins-entity-facts">
+                <GridItem md={ 6 }>
+                    <div>
+                        <span>
```
ryelo: Should this be a `<Level>`?

ryelo: Should this be a` <Level>`?

karelhala: Not really Level will split these two into seperate items, but here we want to have them aligned next to each other.

PR #198, opened 2019-01-15
Show display name and OS release inside inv table

karelhala:


jharting:
@karelhala I think the system name we display in the inventory table should be treated differently. It should not be display_name neither fqdn but rather `systemName = display_name || fqdn`. I.e. display fqdn by default but it a system happens to have display_name defined then show that instead

karelhala:
@jharting yeah that is right, I also added option if neither fqdn or display_name is available to show ID so users can see at least something even if almost no data are available.

I have also moved the data mapping away from api fetch and moved it to reducer where it is much more understandable.

PR #202, opened 2019-01-16
Ensures we only return pagination row once its not of type undefined

AllenBW:
Every no and again, loading inventory table for the first time would throw

### this error :scream:
screen shot 2019-01-16 at 7 55 37 am

Super not cool. It crashes the app.  This pr fixes that


karelhala:
Thanks for the PR, never stumbled apon this error. But better be safe than sorry.

PR #205, opened 2019-01-16
adding skeleton component <?.x.?>

ryelo:


src/PresentationalComponents/Skeleton/Skeleton.js
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        <div className={ skeletonClasses } { ...props }>&nbsp;</div>
+    );
+};
+
+export default Skeleton;
+
+Skeleton.propTypes = {
+    className: propTypes.string,
+    size: propTypes.oneOf([ 'xs', 'sm', 'md', 'lg' ])
```
karelhala: Can you make this to be constant that is exported and consumers can use it?
```JSX
import React from 'react';

import classNames from 'classnames';
export const SkeletonSize = { xs: 'xs', sm: 'sm', md: 'md', lg: 'lg' };
// ...

Skeleton.propTypes = {
    className: propTypes.string,
    size: propTypes.oneOf(Object.values(SkeletonSize))
}

File: src/PresentationalComponents/Skeleton/index.js [hunk 2]
```diff
@@ -0,0 +1 @@
+export { default as Skeleton } from './Skeleton.js';
```
karelhala: `export { default as Skeleton, SkeletonSize } from './Skeleton.js';`

PR #207, opened 2019-01-16
Use lodash flatMap in all files

karelhala:


ryelo:
@karelhala can you fix Travis errors?

PR #221, opened 2019-01-21
make remediation wizard more robust

jharting:


src/SmartComponents/Remediations/steps/ErrorStep.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react';
+import propTypes from 'prop-types';
+
+import './ErrorStep.scss';
+
+export default function ErrorStep ({ message = 'Error occured. Please try again later.' }) {
```
karelhala: Just a nitpick, this can be simplified into
```JSX
const ErrorStep ({ message }) => (
  <p className="ins-c-remediations-error-step">
    { message }
  </p>
);

ErrorStep.propTypes = {
  message: propTypes.string
};

ErrorStep.defaultProps = {
  message: 'Error occured. Please try again later.'
};
export default ErrorStep;
```

jharting: @karelhala is that a simplification given that it adds more chars than the original ;-) ? If we decide to prefer `defaultProps` approach I am down, though.

karelhala: This is somewhat related to when thins get populated. With `defaultProps` you will populated `message` property before construct, with default value in function you will populate the prop on construct time. That is probably the biggest difference.

PR #245, opened 2019-01-23
Allow passing custom props to header column

karelhala:
This will allow changing the width of columns for consumers:
#### Before
selection_021

#### After
selection_022


ryelo:
Are the docs updated to show this too?

karelhala:
@ryelo good point. Gonna add it there as well

PR #247, opened 2019-01-24
Change cursor of ins composed to pointer

karelhala:


ryelo:
@karelhala looks fine. In the past, if it was only 1 line of CSS, we tend to keep it in the same line like:

`.ins-composed-col { cursor: pointer; }`

PR #248, opened 2019-01-24
Add documentation how to use debounce with pagination input

karelhala:


karelhala:
It could be fixed, but that would be really hacky and not that easy as this :cry:

PR #259, opened 2019-01-27
fixes advisorapp fetch, awaits getUser() then calls fetch

AllenBW:
OK ok for whatever reason I wasnt seeing this when reviewing https://github.com/RedHatInsights/insights-frontend-components/pull/253 but ends up `getUser` takes no arguments, but we can await for it, THEN fetch, this solves the original issue...


karelhala:
Good catch! I ment to add `then` but await is much better.

PR #263, opened 2019-01-29
Add credentials include to all fetch requests

karelhala:


jharting:
@karelhala it is not clear to me why this is needed. Based on https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch the default value is same-origin I think is appropriate here. Is this a workaround for a browser bug?

PR #266, opened 2019-01-30
Hide filters if consumer passes items

karelhala:
This is follow up from discussion we had yesterday about hiding filters if consumers (applications) will pass any items (eg try to limit the search).

No change in API, everytime you pass array of items inventory will limit the amount based on current page and per_page option.

Without this change if any aplication asks for more than 100 IDs API will throw error that request is too large and fails.

If we apply this change every application that passes any amount of items will loose the ability to filter inventory table (it was already )

### DIFF
--- a/src/SmartComponents/Inventory/Inventory.js
+++ b/src/SmartComponents/Inventory/Inventory.js
@@ -41,7 +41,12 @@ class InventoryTable extends Component {
             } }>
                 <Card>
                     <CardHeader>
-                        <Filter { ...props } items={ items } filters={ filters } pathPrefix={ pathPrefix } apiBase={ apiBase } />
+                        <Filter { ...props }
+                            hasItems={ items && items.length !== 0 }
+                            filters={ filters }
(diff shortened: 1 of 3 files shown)


src/SmartComponents/Inventory/Inventory.js
```diff
@@ -41,7 +41,7 @@ class InventoryTable extends Component {
             } }>
                 <Card>
                     <CardHeader>
-                        <Filter { ...props } filters={ filters } pathPrefix={ pathPrefix } apiBase={ apiBase } />
+                        <Filter { ...props } items={ items } filters={ filters } pathPrefix={ pathPrefix } apiBase={ apiBase } />
```
AllenBW: So if we're dealing with really large lists of items, maybe this should be a flag/bool rather than passing that really large list yet again somewhere in the app?

karelhala: Good point, gonna add `hasItems` prop.

PR #272, opened 2019-01-31
Use debounce and AbortController when user is filtering hosts

karelhala:


src/SmartComponents/Inventory/Filter/Filter.js
```diff
@@ -105,7 +105,7 @@ class ContextFilter extends Component {
                 filters
             }, () => onRefreshData({ filters: this.props.activeFilters, page: 1 }));
         }
-    }
+    }, 600)
```
AllenBW: should the debounce value be standardized across all uses? ive seen it 800 elsewhere  🤔

karelhala: Good point, I've seen 400 as well. But let's use 800 so we are consistent with pagination as well.

AllenBW: Agreeeeee with this, one pr to change/unite them all would be good...

PR #279, opened 2019-02-05
updating styling for expandable rules card

ryelo:
[RHIADVISOR-407](https://projects.engineering.redhat.com/browse/RHIADVISOR-407)

cc: @kensible

Before:

image

After:
screen shot 2019-02-05 at 10 44 21 am
screen shot 2019-02-05 at 10 44 30 am

This also fixes 
- items inside of `code` tags wouldn't wrap and extend past the edges of the page.
- Removing unnecessary Grid components when Stacks should be used
- redundant styling

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6816,7 +6816,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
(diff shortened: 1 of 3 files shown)


dkuc:
The table looks odd to me being stretched to the entire width. I guess that's how legacy does it though, so maybe its not an issue.

ryelo:
@dkuc I can certainly make it smaller and it would take just a few seconds, so if anyone else agrees, it'll be an easy fix.

AllenBW:
Wait wait, so we're ok with blowing away author table style? I'm 😕

ryelo:
@AllenBW we blew it away in classic because the author table styling is essentially non existent as well as the data (mainly the headers) being illegible being so close together.

PR #289, opened 2019-02-07
Updating toolbar to support results section

ryelo:
image

### DIFF
--- a/src/PresentationalComponents/TableToolbar/TableToolbar.js
+++ b/src/PresentationalComponents/TableToolbar/TableToolbar.js
@@ -5,20 +5,20 @@ import classNames from 'classnames';
 
 import './TableToolbar.scss';
 
+function generateResults(results) {
+    if (results > 1 || results < 1) {
+        return (`${results} Results`);
+    } else {
+        return (`${results} Result`);
+    }
+}
+
 const TableToolbar = ({ results, className, children, ...props }) => {
 
     const tableToolbarClasses = classNames(
         'ins-c-table__toolbar',
         className
     );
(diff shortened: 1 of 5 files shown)


src/PresentationalComponents/TableToolbar/TableToolbar.js
```diff
@@ -1,25 +1,42 @@
...
 
-const TableToolbar = ({ className, children, ...props }) => {
+const TableToolbar = ({ results, className, children, ...props }) => {
 
     const tableToolbarClasses = classNames(
         'ins-c-table__toolbar',
         className
     );
 
+    function generateResults(results) {
```
karelhala: I might have been too fast with the approve button. Can we move this function above `TableToolbar` component? So it is not recreated everytime we re render table toolbar?

src/PresentationalComponents/TableToolbar/TableToolbar.test.js
```diff
@@ -8,4 +8,19 @@ describe('TableToolbar component', () => {
         const wrapper = shallow(<TableToolbar>Some</TableToolbar>)
         expect(toJson(wrapper)).toMatchSnapshot();
     });
+
+    it('should render with results of 0', () => {
+        const wrapper = shallow(<TableToolbar result={ 0 }>Some</TableToolbar>)
```
karelhala: Typo in all tests `result` -> `results`

ryelo: 🤦‍♂️

PR #290, opened 2019-02-08
Vulnerability: rewrite CVE table to PF4

jiridostal:
* Use PF4 for table
* Provide parameter to set default sort
* Move pagination to bottom
* Reverse initial sort direction when clicking a column

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -233,6 +233,29 @@
         }
       }
     },
+    "@emotion/cache": {
+      "version": "10.0.7",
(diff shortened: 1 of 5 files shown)


karelhala:
Tested with empty data and table looks good! We will have to work on empty state and plagination placement, but let's wait for https://github.com/RedHatInsights/insights-frontend-components/pull/282 to be closed before we dive into it.

screenshot from 2019-02-08 13-27-35

PR #292, opened 2019-02-08
Wizard.js - Allow confirmAction prop to override confirm button text

kruai:
Add an optional override to the "Confirm" button's text.
This was created so that a Wizard created in Remediations could have the button say "Save" instead.

### DIFF
--- a/src/PresentationalComponents/Wizard/Wizard.js
+++ b/src/PresentationalComponents/Wizard/Wizard.js
@@ -42,7 +42,7 @@ class Wizard extends Component {
 
     render() {
 
-        const { isLarge, title, className, isOpen, isValidated, ...props } = this.props;
+        const { isLarge, title, className, isOpen, isValidated, confirmAction, ...props } = this.props;
 
         const wizardClasses = classNames(
             'ins-c-wizard',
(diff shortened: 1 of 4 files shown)


src/PresentationalComponents/Wizard/Wizard.js
```diff
@@ -72,7 +72,7 @@ class Wizard extends Component {
                     variant="primary"
                     isDisabled={ !isValidated }
                     onClick={ () => this.handleOnClose(true) }>
-                        Confirm
+                    { this.props.confirmAction }
```
karelhala: Please do not reference confirm action trough `this.props` but rather add `confirmAction` to https://github.com/RedHatInsights/insights-frontend-components/blob/72aa2d227bc92230f82397a7b90ecd2af51f84d6/src/PresentationalComponents/Wizard/Wizard.js#L45

karelhala:
Please add tests to this functionality and run `npm t -- -u` to update snapshots as well.

src/PresentationalComponents/Wizard/Wizard.js
kruai: Will do. Is this just for readability, or is there a performance difference as well?

karelhala: Mostly for readability. There might be a slight performance difference, but that won't be noticable either way.

PR #295, opened 2019-02-13
Add warn msg about dropdown deprecation and add tests

karelhala:
RHIPLAT2-446
partial RHIPLAT2-434


ryelo:
@karelhala looks like some linting errors

PR #297, opened 2019-02-13
add Remediation button to Insights/Vulnerability system detail view

jharting:
screenshot_20190213_171249
screenshot_20190213_171432

### DIFF
--- a/src/SmartComponents/Inventory/applications/advisor/Advisor.js
+++ b/src/SmartComponents/Inventory/applications/advisor/Advisor.js
@@ -9,6 +9,7 @@ import ExpandableRulesCard from './ExpandableRulesCard';
 import { Card, CardBody, CardHeader } from '@patternfly/react-core';
 import { CommentSlashIcon, FrownOpenIcon } from '@patternfly/react-icons';
 import { withRouter } from 'react-router-dom';
+import RemediationButton from '../../../Remediations/RemediationButton';
 
 import './advisor.scss';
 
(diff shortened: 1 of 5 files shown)


src/SmartComponents/Inventory/applications/advisor/Advisor.js
```diff
@@ -82,17 +93,21 @@ class InventoryRuleList extends Component {
     }
 
     buildRuleCards = () => {
-        const { activeReports, expanded, kbaDetails } = this.state;
+        const { activeReports, expanded, kbaDetails, remediation } = this.state;
 
         return (
             <Fragment>
-                <div className="pf-u-display-flex pf-u-flex-row-reverse">
+                <div className="pf-u-display-flex pf-u-flex-row-reverse ins-c-advisor-system-actions">
```
ryelo: Can we just use the `<Level>` component here instead?

jharting: @ryelo done

AllenBW:
Ok I missed this, but it pushes the `collapse/expand` button to the left which does not follow the current mocks... furthermore, if this IS the new direction, lines 5 and 6, the two patternfly styles that are no longer used should be removed

jharting:
@AllenBW wait, this is what I saw (and still can reproduce) when I run Advisor with `insights-frontend-components@3.33.64` i.e. just before my changes:

screenshot_20190215_083237

Based on that I assumed the collapse/expand button is moving to the left and used the room on the right for my button.

@AllenBW can you point me to the current set of mocks for Insights?

AllenBW:
Ah! I guess the PF version updated, that second reverse direction flex class needed to be reworded when we updated... Atm we're using screen shots of beta, which has this on the right 😬

PR #302, opened 2019-02-14
Allow inventory table to overflow card

karelhala:
screenshot from 2019-02-14 13-29-31

RHIPLAT2-320

### DIFF
--- a/src/SmartComponents/Inventory/Inventory.js
+++ b/src/SmartComponents/Inventory/Inventory.js
@@ -49,7 +49,7 @@ class InventoryTable extends Component {
                             apiBase={ apiBase }
                         />
                     </CardHeader>
-                    <CardBody>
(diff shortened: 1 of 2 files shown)


ryelo:
What's the screen size for this screenshot?

karelhala:
@ryelo I had to improvise because basic inventory table is not wide enough. But because we allow additional cells to be added we have to allow this scroll for instance in compliance detail they have 6 cells for each row and on screen smaller than 930px table overflows it's parent and data are not readable. However if the screen size is small it is rendered completely differently.

ryelo:
@karelhala So this is a temporary fix? This should be brought up with the PF team too?

karelhala:
@ryelo yeah temporary fix and issue has been already opened https://github.com/patternfly/patternfly-next/issues/1418

ryelo:
Sweet!

PR #303, opened 2019-02-14
adding empty table state <?.x.?>

ryelo:
screen shot 2019-02-14 at 8 53 37 am

### DIFF
--- a/src/PresentationalComponents/EmptyTable/EmptyTable.js
+++ b/src/PresentationalComponents/EmptyTable/EmptyTable.js
@@ -1,0 +1,27 @@
+import React, { Fragment } from 'react';
+import propTypes from 'prop-types';
+
+import classNames from 'classnames';
+
+import './EmptyTable.scss';
+
+const EmptyTable = ({ centered, className, children, ...props }) => {
+
+    const emptyTableClasses = classNames(
+        'ins-c-table__empty',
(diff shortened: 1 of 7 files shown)


src/PresentationalComponents/EmptyTable/EmptyTable.js
```diff
@@ -0,0 +1,27 @@
...
+const EmptyTable = ({ centered, className, children, ...props }) => {
+
+    const emptyTableClasses = classNames(
+        'ins-c-table__empty',
+        { [`is-centered`]: centered },
+        className
+    );
+
+    return (
+        <div className={ emptyTableClasses } { ...props }> { children }</div>
```
karelhala: Can we use [patternfly-4/components/emptystate](http://patternfly-react.surge.sh/patternfly-4/components/emptystate) here? So we are a bit more aligned to PF4?

ryelo: I toyed around with that idea, but it looks weird. Plus, I don't think that this would be a place that it would be needed. They could always add an empty state inside of this, but we just need to provide styling to look like a table.

karelhala: Oh, okay. Makes sense, good to merge then.

PR #310, opened 2019-02-18
Allow inventory to be paginated by application

karelhala:
No UI changes

RHIPLAT2-464

* [x] test inside application
* [x] test inside host inventory


jiridostal:
Works very well! Only a note - 

- now as there's a possibility to pass custom pagination, it should also be possible to sort by application. Combining application pagination and Inventory "sort loaded items..." does not produce convenient result.

karelhala:
@jiridostal good point! I'll create new issue to accommodate it.

PR #311, opened 2019-02-18
Shield Component (former ImpactBadge Component)

janelznic:
You can overview and merge. cc @karelhala


janelznic:
The CI Travis fail seems to be related to battery component not to my changes. Should I do anything else?

package.json
```diff
@@ -13,8 +13,8 @@
     "Utilities/"
   ],
   "dependencies": {
-    "@patternfly/react-core": "^1.51.4",
-    "@patternfly/react-icons": "^2.8.0",
+    "@patternfly/react-core": "^1.52.0",
```
karelhala: Is there a reason to bump `@patternfly/react-core` and `@patternfly/react-icons` dependencies?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
+import { QuestionIcon, SecurityIcon } from '@patternfly/react-icons';
+import { Tooltip } from '@patternfly/react-core';
+import propTypes from 'prop-types';
+import React from 'react';
+
+class ImpactBadge extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.colorList = {
```
karelhala: Can you move this either above the component or to special const file? So we don't recreate these values everytime component is created?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+            default: 'var(--pf-global--Color--200)', // grey
```
karelhala: Can you move this to styles file? I don't really like using inline styles, but you can use something like `inc-c-critical-badge`

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+                color: this.colorList.warning
+            },
+            Low: {
+                title: 'Low',
+                color: this.colorList.default
+            }
+        };
+    }
+
+    getColoredBadgeByImpact(impact) {
```
karelhala: No need to pass impact, it's in props.

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+        const iconSize = this.props.size || 'md';
```
karelhala: Can you setup default prop `size` to md instead of this shortcutting?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+        let badge;
+
+        if (Object.keys(this.impactList).includes(impact)) {
+            badge = {
```
karelhala: No need to assign to variable if you are not planning to do something with it, just return it.

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+                color: this.colorList.default
+            }
+        };
+    }
+
+    getColoredBadgeByImpact(impact) {
+        const iconSize = this.props.size || 'md';
+        let badge;
+
+        if (Object.keys(this.impactList).includes(impact)) {
```
karelhala: This is overcomplicated way of `this.impactList.hasOwnProperty(impact)`

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        return badge;
+    }
+
+    render() {
+        const badge = this.getColoredBadgeByImpact(this.props.impact);
+        return (
+            <React.Fragment>
+                <div>
```
ryelo: Can we eliminate this `<div>`? It's not needed

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+        this.colorList = {
+            default: 'var(--pf-global--Color--200)', // grey
+            danger: 'var(--pf-global--danger-color--100)',
+            warning: 'var(--pf-global--warning-color--100)',
+            orange: '#ec7a08' // orange
+        };
+
+        /* I'm not shure that the list of impacts is ordered correctly by relevancy */
+        this.impactList = {
+            Critical: {
```
ryelo: Can we safely assume these values will always be uppercase?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+
+        return badge;
+    }
+
+    render() {
+        const badge = this.getColoredBadgeByImpact(this.props.impact);
+        return (
+            <React.Fragment>
+                <div>
+                    <Tooltip position="right" content={ <div>Impact: { badge.title }</div> }>
```
ryelo: Typically, we stick with having the tooltip on the top. Can we change this and then allow a prop to change the position?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
```diff
@@ -0,0 +1,85 @@
...
+                        { badge.icon }
+                    </Tooltip>
+                </div>
+            </React.Fragment>
+        );
+    }
+}
+
+ImpactBadge.propTypes = {
```
ryelo: Should we add some default props?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
ryelo: Can we also change this so it doesn't have to say "Impact:"?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
janelznic: It's needed to wrap the `<Tooltip />` component as a near relative element because positioning.

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
janelznic: cc @jiridostal Please can you answer and make a conclusion?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
janelznic: What we would to say instead of _impact_ word?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
jiridostal: As Vulnerability engine stores impacts as a dictionary of uppercase values, we can assume it will either be uppercase word or empty string.

package.json
janelznic: Yes, we are using features like `<Tooltip>` from new version.

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
janelznic: Where I should to place this special const file and name it (including class name) in file structure, please? Is there any rule for it? Or existing example?

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
janelznic: I'm just using color property from https://github.com/patternfly/patternfly-react/tree/master/packages/react-icons. Should I use the external CSS definition anyway?

package.json
karelhala: Can you identify which packages are required to make this running? `@patternfly/react-core`, `@patternfly/react-icons` and `react-dom`? Anything else? I will create new PR that updates these packages so we have slightly better consistency.

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
karelhala: I don't think so, but you can create new file next to this one called `consts.js` and export these variables there.

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
karelhala: Css definitions would be slightly better, but I guess you can keep it this way for simplicity.

package.json
janelznic: OK, please make a new PR with these changes:
```
"@patternfly/react-core": "^1.52.0",
"@patternfly/react-icons": "^2.10.0",
"react-dom": "^16.7.0",

```

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
ryelo: [Tooltip Docs](https://patternfly-react.surge.sh/patternfly-4/components/tooltip) don't have theirs wrapped in a div.

You can see my example [here](https://stackblitz.com/edit/react-vbjycm)

ryelo:
@janelznic I think this component needs to be abstracted. If the "ImpactBadge" is local enough to Vulnerabilities, it should probably live there.

IMO, if we want to contribute this component so that other apps can use it (I believe advisor wanted to use it), we need to abstract this from `ImpactBadge` to `Shield` and then not have it so specific to Impact, much like we are doing with the battery. You can supply a number or a severity to the component and then are able to label it as anything you would like.

@karelhala Thoughts?

package.json
karelhala: Sure thing! New PR has been opened https://github.com/RedHatInsights/insights-frontend-components/pull/317

package.json
```diff
@@ -30,7 +30,7 @@
     "prop-types": "^15.6.2",
     "react": "^16.5.1",
     "react-apollo": "^2.3.3",
-    "react-dom": "^16.5.1",
+    "react-dom": "^16.7.0",
```
karelhala: React dom has been updated to `16.8.2` in master

karelhala:
@ryelo I really like the idea of adding new component called `Shield` with different impact values. @janelznic you can keep your mapping, but you should also add some general values
```JS

impactList = {
  // defined as before
}

impactList.maximum = impactList.Critical;
// continue with some generic values
```

src/PresentationalComponents/ImpactBadge/ImpactBadge.js
janelznic: Added prop `tooltipPrefix `(default is empty)

karelhala:
@jiridostal any suggestions?
@ryelo can you re-review?

src/PresentationalComponents/Shield/Shield.js
```diff
@@ -0,0 +1,63 @@
...
+                ) }
+            </React.Fragment>
+        );
+    }
+}
+
+Shield.defaultProps = {
+    impact: 'N/A',
+    hasTooltip: false,
+    tooltipPosition: 'right',
```
ryelo: This needs to be defaulted to top

src/PresentationalComponents/Shield/consts.js
```diff
@@ -0,0 +1,39 @@
+export const colorList = {
+    default: 'var(--pf-global--Color--200)', // grey
+    danger: 'var(--pf-global--danger-color--100)',
+    warning: 'var(--pf-global--warning-color--100)',
+    orange: '#ec7a08' // orange
+};
+
+/* I'm not shure that the list of impacts is ordered correctly by relevancy */
+export const impactList = {
```
ryelo: Can we also support 4 levels numerically?
```
4: Critical
3: High
2: Medium
1. Low
```

src/PresentationalComponents/Shield/Shield.js
```diff
@@ -0,0 +1,63 @@
...
+
+class Shield extends React.Component {
+    constructor(props) {
+        super(props);
+    }
+
+    getColoredBadgeByImpact() {
+        if (impactList.hasOwnProperty(this.props.impact)) {
+            return {
+                icon: <SecurityIcon size={ this.props.size } color={ impactList[this.props.impact].color } />,
```
ryelo: We have to have some kind of `aria-label` that describes what the Badge is representing.

This means title needs to be a required field.

It should display like `aria-label = 'Impact High'`

src/PresentationalComponents/Shield/consts.js
janelznic: How? Like this?
`this.impactList.numerically = { 1: 'Low', 2: 'Medium', 3: 'High', 4: 'Critical' }`

ryelo: Yes. That's perfect.

src/PresentationalComponents/Shield/Shield.js
```diff
@@ -0,0 +1,76 @@
...
+import { Tooltip } from '@patternfly/react-core';
+import propTypes from 'prop-types';
+import React from 'react';
+import { impactList, colorList } from './consts';
+
+class Shield extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.title = this.props.tooltipPrefix + (this.props.title || this.getColoredBadgeByImpact().title);
```
karelhala: Assigning to attributes is generally not a good idea. With rhis approach it means that whever consumer changes title it will not get re-rendered. Can you move this to render and getColoredBadgeByImpact please?

PR #314, opened 2019-02-20
Add warn msg not to use breadcrumbs anymore

karelhala:


src/PresentationalComponents/Breadcrumbs/ConnectedBreadcrumbs.js
```diff
@@ -7,6 +7,7 @@ class ConnectedBreadcrumbs extends Component {
         super(props);
         this.onNavigate = this.onNavigate.bind(this);
         this.calculateBreadcrumbs = this.calculateBreadcrumbs.bind(this);
+        console.warn('This component will be removed in future release, do not use it anymore!');
```
ryelo: Can we come up with a date for these?

karelhala: Sure I don't really thing anybody is using it anyways.

PR #316, opened 2019-02-20
Update pf4 react table to newest version

karelhala:


jiridostal:
I need to get back to this PR. The code change in loading state causes table to shrink when loading 
image. To me it seems that colSpan does not remove "connected" columns but displays them at the end of the table.

karelhala:
@jiridostal looks like the the colspan was badly calculated https://github.com/RedHatInsights/insights-frontend-components/blob/master/src/SmartComponents/Inventory/applications/vulnerabilities/VulnerabilitiesCveTable.js#L87

jiridostal:
image
Yes but still - the rest of columns remain there even if colspan is set

PR #319, opened 2019-02-21
Correct loading for inventory detail

karelhala:
This fixes bug that when inventory detail was actually loaded, but no app was visible it showed loading skeleton that can never be finished.


karelhala:
@dLobatog the difference is that the top part (where this loader was) is loaded trough chrome - async loaded so that's why you have to load it trough chrome.

PR #320, opened 2019-02-22
Add new component skeleton table

karelhala:


ryelo:
@karelhala we recently had discussions about loading states. From what we talked about, the skeleton is only supposed to be used if you know exactly what data is going to be there. In this scenario, we don't know the rows or columns, so there should only be a spinner inside an empty table

src/PresentationalComponents/Pagination/PaginationNav.js
```diff
@@ -45,7 +45,7 @@ const PaginationNav = ({
                 min="1"
                 data-action="set-page"
                 max={ amountOfPages }
-                style={ { width: `${lastPage.toString().length}rem` } }
+                style={ { width: `${lastPage.toString().length}rem`, minWidth: '3rem' } }
```
AllenBW: yes please 🤗

karelhala:
> From what we talked about, the skeleton is only supposed to be used if you know exactly what data is going to be there. In this scenario, we don't know the rows or columns, so there should only be a spinner inside an empty table

Well this skeleton table is for tables you know how many columns they have. So you kinda know how big each column will be. And you also know how many items you are fetching

ryelo:
@karelhala Yeah, that makes sense. I think this was more for the rows. If we show 10 rows and only one shows up, that will be breaking the (future) rules. We would use a spinner in that case.

karelhala:
@ryelo oh, okay that makes sense. What would be the optimal path here? Wait until we figure out spinner? Or merge this and update once we get the spinner and such out?

ryelo:
@karelhala Spinner is supposed to be implemented in react by March 22. I can approve this for now, or we can wait until the spinner is implemented. Doesn't matter to me.

karelhala:
@ryelo well we can merge this one and deprecate/update later on when PF implements spinner.

ryelo:
@karelhala WFM!

src/PresentationalComponents/SkeletonTable/SkeletonTable.js
```diff
@@ -0,0 +1,40 @@
...
+        return (
+            <Table cells={ this.createColumns() } rows={ this.createRows() } aria-label="Loading">
+                <TableHeader />
+                <TableBody />
+            </Table>
+        );
+    }
+}
+
+SkeletonTable.propTypes = {
```
ryelo: Can we give the user an option to pass in their headers?

src/PresentationalComponents/SkeletonTable/SkeletonTable.test.js
```diff
@@ -0,0 +1,16 @@
...
+import SkeletonTable from './SkeletonTable';
+
+describe('SkeletonTable component', () => {
+    it('should render correctly', () => {
+        const wrapper = shallow(<SkeletonTable colSize={5} rowSize={15}/>)
+        expect(toJson(wrapper)).toMatchSnapshot();
+    })
+
+    it('should render correctly without rows', () => {
+        const wrapper = shallow(<SkeletonTable colSize={5} rowSize={15} />)
```
ryelo: Do you want to remove the `rowSize` prop for this one?

PR #328, opened 2019-02-27
Systems Exposed Table - Redesigned table to new wireframe design

janelznic:


src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
```diff
@@ -18,13 +10,22 @@ export function createCveListBySystem({ isLoading, ...rest }) {
...
+                        hasTooltip={ true }
+                        tooltipPosition={ 'right' }
+                        key={ row.id.toString() }
+                    />,
                     row.attributes.synopsis,
-                    <span key={ `title-${row.id}` } title={ row.attributes.description }>{ row.attributes.description.substr(0, 199) }...</span>,
+                    <LongTextTooltip
+                        content={ row.attributes.description }
+                        maxLength={ 200 }
+                        tooltipPosition={ 'top' }
```
jiridostal: Can we make sure it's consistent with https://github.com/RedHatInsights/vulnerability-ui/blob/master/src/Helpers/VulnerabilitiesHelper.js#L19-L24

src/SmartComponents/Inventory/applications/vulnerabilities/vulnerabilities.scss
```diff
@@ -28,9 +28,16 @@
...
 .number-of-results {
   display: inline-block;
   min-width: 55px;
   text-align: right;
   margin-right: 5px;
 }
+
+.table-header-icon {
+  svg {
+    margin-right: 5px;
```
ryelo: Please don't use `px` values. Use the px to rem mixin located in the Utilities

src/SmartComponents/Inventory/applications/vulnerabilities/vulnerabilities.scss
```diff
@@ -28,9 +28,16 @@
 #download-report {
```
ryelo: Try to avoid styling using ID tags due to specificity levels

src/SmartComponents/Inventory/applications/vulnerabilities/Vulnerabilities.js
```diff
@@ -3,17 +3,36 @@ import { connect } from 'react-redux';
...
-    { title: 'Publish date', key: 'public_date', transforms: [ sortable, cellWidth(10) ]}
+    {
+        title: (
+            <React.Fragment>
+                <Tooltip
+                    position="top"
+                    content={ <div>{ cvssBaseDescription }</div> }
+                >
+                    <span className="table-header-icon">
+                        <InfoCircleIcon arial-label={ cvssBaseDescription } aria-hidden="false" />
```
ryelo: Why are you using `aria-hidden: false`?

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
```diff
@@ -18,13 +10,21 @@ export function createCveListBySystem({ isLoading, ...rest }) {
             data: data.map(row => ({
                 id: row.id,
                 cells: [
-                    <span key={ row.id }>
-                        <SecurityIcon size="md" color={ colorToImpact[row.attributes.impact] } />
-                    </span>,
+                    <Shield
+                        impact={ row.attributes.impact }
+                        hasTooltip={ true }
+                        tooltipPosition={ 'right' }
```
ryelo: Tooltip needs to be on the top

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
```diff
@@ -18,13 +10,21 @@ export function createCveListBySystem({ isLoading, ...rest }) {
...
+                        key={ row.id.toString() }
+                    />,
                     row.attributes.synopsis,
-                    <span key={ `title-${row.id}` } title={ row.attributes.description }>{ row.attributes.description.substr(0, 199) }...</span>,
+                    <LongTextTooltip
+                        content={ row.attributes.description }
+                        tooltipMaxWidth={ '50vw' }
```
ryelo: Why are we styling using the viewport?

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
```diff
@@ -18,13 +10,21 @@ export function createCveListBySystem({ isLoading, ...rest }) {
...
+                        entryDelay="1200"
```
ryelo: Let's keep the delay for tooltips consistent with Patternfly's specs

src/SmartComponents/Inventory/applications/vulnerabilities/Vulnerabilities.js
```diff
@@ -3,17 +3,36 @@ import { connect } from 'react-redux';
...
-    { title: 'CVSS Base Score', key: 'cvss_score', transforms: [ sortable, cellWidth(10) ]},
-    { title: 'Publish date', key: 'public_date', transforms: [ sortable, cellWidth(10) ]}
+    {
+        title: (
+            <React.Fragment>
+                <Tooltip
+                    position="top"
+                    content={ <div>{ cvssBaseDescription }</div> }
+                >
+                    <span className="table-header-icon">
```
ryelo: Is there a reason to wrap this in a `span` tag instead of just using the icon?

src/SmartComponents/Inventory/applications/vulnerabilities/vulnerabilities.scss
janelznic: It's not the (my) code related to this pull request. Please create a new issue.

src/SmartComponents/Inventory/applications/vulnerabilities/Vulnerabilities.js
janelznic: Because the default value of `aria-hidden` attribute is `true`. And we need to get the `aria-label` readable for screen readers.

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
janelznic: I have an assignment to be right because it is the first column in the table. When the window size width is small, the `top` position of tooltip may not be displayed correctly:
screenshot 2019-02-28 08 31 29

Should I use `top` position anyway?

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
janelznic: What other solution do you suggest? This is the default size for the `LongTextTooltip` component.

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
janelznic: @jiridostal required that. Jirko, please leave a comment.

src/SmartComponents/Inventory/applications/vulnerabilities/Vulnerabilities.js
janelznic: No, actually it's not. Fixed.

src/SmartComponents/Inventory/applications/vulnerabilities/vulnerabilities.scss
janelznic: OK, updated to `.3rem`.

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
jiridostal: If the entry delay is on default value in this case tooltips get very annoying as there is large hover area in each row of table and the tooltip content is usually very long. Scrolling on the page is very uncomfortable with the default entry delay - so we may need UXD support on this, however keeping the default pretty much ruins the experience.

ryelo: @jiridostal Makes sense to me. Let's make a note to get UXD on this, but it'll be fine for now.

src/SmartComponents/Inventory/applications/vulnerabilities/DataMapper.js
ryelo: 🤔 This may be a problem with the tooltip component from the PF side. Let me investigate further. Make a note and we'll merge it for now.

src/SmartComponents/Inventory/applications/vulnerabilities/Vulnerabilities.js
ryelo: Looking at the icons for PF, they aren't assigned a value for `aria-hidden` meaning that screen readers should pick it up because of the label.

janelznic: Unfortunately that's not correct. When I remove the `aria-hidden="true"` attribute:
`<InfoCircleIcon className={ 'table-header-icon' } arial-label={ cvssBaseDescription } />`

Patterfly adding `aria-hidden="false"`. Really:
screenshot 2019-02-28 14 24 57

ryelo: Ah yes. I see now, that's because you're already describing the icon with "CVSS Base Score" if it was `aria-hidden=false`, then screen readers would read it twice.

src/SmartComponents/Inventory/applications/vulnerabilities/vulnerabilities.scss
```diff
@@ -37,7 +37,5 @@
 }
 
 .table-header-icon {
-  svg {
-    margin-right: 5px;
-  }
+  margin-right: .3rem;
```
ryelo: Please import `Utilities/_mixins.scss`

then do:
` .table-header-icon svg { @include rem('margin-right', 5px); }`

src/Utilities/helpers.js
```diff
@@ -28,14 +28,20 @@ export function parseCvssScore(cvssV2, cvssV3) {
...
         (cvssV2 && `${parseFloat(cvssV2).toFixed(1)} (v2)`) ||
         'N/A'
-    )
+    );
+}
+
+export function processDate(dateString) {
+    const date = new Date(dateString);
+    const dateFormat = require('dateformat');
+    return (date instanceof Date && !isNaN(date) && dateFormat(date, 'd mmm yyyy')) || 'N/A';
```
katierik: Question - Does "d mmm yyyy" handle "19 Feb 2019" (aka dd mmm yyyy)? Does it go "07" or "7"?

ryelo: I believe `d mmm yyyy` is going to eliminate the leading zero.

katierik: Double checked with our content person and it should be 07 Feb 2019 - hopefully a quick fix? :)

jiridostal: Fixed

ryelo:
@katierik Everything good on your end now?

jiridostal:
I've added one more simple change - should not block on merging.

PR #331, opened 2019-02-27
VMAAS-348 Download Button Tooltip

janelznic:


ryelo:
What's the purpose for putting a tooltip on the dropdown? Typically, tooltips are not used on form elements such as dropdowns.

janelznic:
> What's the purpose for putting a tooltip on the dropdown? Typically, tooltips are not used on form elements such as dropdowns because they're interactive elements that need to be clicked. If there is a necessary need for this, let's get UXD involved to figure out the design.

@ryelo: Please can you contact the task reporter Martin Kourim <mkourim@redhat.com>?
It's in the assignment: https://projects.engineering.redhat.com/browse/VMAAS-348

mkoura:
@janelznic @ryelo It's not obvious that the download icon is dropdown and as there's no tooltip, the user don't know what will happen when he click the icon. Either it should be obvious the icon is dropdown, i.e. safe to click, or tooltip should be present. This is just my opinion from user pov, let's get UXD involved.

ryelo:
@mkoura I think this is a tricky scenario, but this follows the pattern for [Dropdown - Icon Only](https://patternfly-react.surge.sh/patternfly-4/components/dropdown). I think if it does create confusion, we should have UXD involved.

ryelo:
An update on this:
> there was much debate about this. There is also no arrow for the kebabs or the info icon or any of the icons in the masthead all of which open menus. So we decided to eliminate arrows for consistency rather than having half with and half without even though they all open some type of menu or popover.

From UXD it sounds like we're just going to keep it the way it is and not add a tooltip or an arrow

mkoura:
FYI this was also brought up by a customer during pilot - https://bugzilla.redhat.com/show_bug.cgi?id=1687866

ryelo:
@mkoura Thanks! I'll bring it up with UXD and see if they'll reconsider.

ryelo:
@mkoura We're going to reopen this. UXD says we should be fine to temporarily do a top tooltip that says

> "Download a  CSV or JSON file"

jiridostal:
Closing as current mock-ups do not use icon for downloading a report. I will create another PR which accomplishes it.

dajohnso:
@jiridostal , can you please update the BZ with the latest around this one since its moving around, THX

https://bugzilla.redhat.com/show_bug.cgi?id=1687866

PR #333, opened 2019-02-27
Shield component props inheritance to Tooltip

janelznic:
cc @jiridostal Please check.


janelznic:
@ryelo: OK, fixed and passed by CI.

PR #335, opened 2019-02-28
patternfly-next > patternfly

ryelo:
This gets our core PF/react PF packages into a more stable point

If an app is using the `Utilities/foo` directory and importing colors, they will need to update their `@patternfly/patternfly-next` to `@patternfly/patternfly`


ryelo:
Tagging this as on-hold until we merge other packages

PR #337, opened 2019-02-28
Fix pagination total of system rules table

akofink:
Follow up to https://github.com/RedHatInsights/insights-frontend-components/pull/327


dLobatog:
Thanks @akofink ~!

PR #349, opened 2019-03-08
Adding ability to show selected count in TableToolbar <?.?.x>

ryelo:
Adds ability to show selected count, results count, or both

Also updated the colors to closer match pagination 

screen shot 2019-03-08 at 1 22 56 pm

### DIFF
--- a/src/PresentationalComponents/TableToolbar/TableToolbar.js
+++ b/src/PresentationalComponents/TableToolbar/TableToolbar.js
@@ -5,7 +5,7 @@ import classNames from 'classnames';
 
 import './TableToolbar.scss';
 
-function generateResults(results) {
+function generateCount(results) {
     if (results > 1 || results < 1) {
         return (`${results} Results`);
     } else {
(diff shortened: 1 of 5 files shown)


src/PresentationalComponents/TableToolbar/TableToolbar.js
```diff
@@ -40,7 +45,8 @@ TableToolbar.propTypes = {
     isFooter: propTypes.bool,
     results: propTypes.number,
     children: propTypes.any,
-    className: propTypes.string
+    className: propTypes.string,
+    selected: propTypes.number
```
karelhala: This should be number. Or `oneOfType([ propTypes.number, propTypes.string])`

ryelo: @karelhala what should be number? `selected` is a number

karelhala: Yeah, my bad. I saw string.

PR #352, opened 2019-03-13
Fixes RHICOMPL-106: remove extra whitespace in table toolbar

akofink:
This completely removes the number of results at the top of the table. I'm not sure why that was part of the `Filter` component in the first place? The same information is still at the bottom of the table. It also aligns the download button with the side of the table and fixes the line above the table.

Before:

image

After:

image

I'm not sure what happened to the 'Remediate with Ansible' button. The patternfly docs suggest it should be pushed all the way to the right... https://patternfly-react.surge.sh/patternfly-4/layouts/level https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/SystemsTable.js#L57

### DIFF
--- a/src/PresentationalComponents/TableToolbar/TableToolbar.scss
+++ b/src/PresentationalComponents/TableToolbar/TableToolbar.scss
@@ -2,7 +2,7 @@
 
 :root {
     --ins-c-table-toolbar--PaddingVertical: 16px;
-    --ins-c-table-toolbar--PaddingHorizontal: 32px;
+    --ins-c-table-toolbar--PaddingHorizontal: 0px;
 }
 
 .ins-c-table__toolbar, .ins-c-table__toolbar-results {
--- a/src/SmartComponents/Inventory/Filter/Filter.js
+++ b/src/SmartComponents/Inventory/Filter/Filter.js
@@ -120,7 +120,7 @@ class ContextFilter extends Component {
         const filteredColumns = columns && columns.filter(column => !column.isTime).map(this.textualFilter);
         const placeholder = filterByString || (filteredColumns && filteredColumns.length > 0 && filteredColumns[0].title);
         return (
-            <Grid guttter="sm" className="ins-inventory-filters">
+            <Grid gutter="sm" className="ins-inventory-filters">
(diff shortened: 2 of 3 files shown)


src/PresentationalComponents/TableToolbar/TableToolbar.scss
```diff
@@ -2,7 +2,7 @@
 
 :root {
     --ins-c-table-toolbar--PaddingVertical: 16px;
-    --ins-c-table-toolbar--PaddingHorizontal: 32px;
+    --ins-c-table-toolbar--PaddingHorizontal: 0px;
```
ryelo: This statement is needed for the toolbar to work as intended. If there is other padding, it is coming from somewhere else. 
Screen Shot 2019-03-13 at 3 50 25 PM

ryelo:
I think this is due to the Card, the table should be extracted from the card and the TableToolbar should just sit on top of the Table

src/PresentationalComponents/TableToolbar/TableToolbar.scss
akofink: Show me this example in relation to the table. This issue was raised because the toolbar and table left/right edges weren't aligned. Perhaps instead it'd be better to add padding to the table?

ryelo: Sure thing:

Screen Shot 2019-03-14 at 8 33 22 AM

ryelo: If you're doing:
```
<TableToolbar>
    <Toolbar>
.
.
.
```
That is incorrect. The `TableToolbar` takes place of the `Toolbar` component. https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/tableToolbar.md

akofink: Ok, thanks!

akofink:
Closing in favor of https://github.com/RedHatInsights/compliance-frontend/pull/95 Thanks @ryelo!

PR #355, opened 2019-03-19
Adjustments to Vulnerability table toolbar

jiridostal:
Before: image

After:
image
There is still Filter component missing, but it should not block this PR.


karelhala:
This changes a lot of things, can we get after screenshot so we know how it would look like after merging this PR?

jiridostal:
I've updated initial commit

PR #357, opened 2019-03-20
Updates url apis per the fe cloud migration guide

AllenBW:
---
name: Updates url apis per the fe cloud migration guide
about: Fixing a bug
title: "Updates url apis per the fe cloud migration guide"
labels: bug
assignees: ''karelhala"

---

as per documentation https://docs.google.com/document/d/1Ku_MELkCjhY5ggwyQ7aawftiLMoKlqdpdsRPPSr8nyI/edit#heading=h.wk0oy98zxmyh

doesn't supersede https://github.com/RedHatInsights/insights-frontend-components/pull/356

PR #358, opened 2019-03-20
Use new API for inventory

karelhala:


karelhala:
Closed because this is addressed in https://github.com/RedHatInsights/insights-frontend-components/pull/357

PR #361, opened 2019-03-25
Insights api changed, report.resoution.has_playbook now, fixes kba fetch

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-503

this fixes the kba fetch issue, the thing that happened because we moved/moving to cloud.redhat

also fixes the move of the has_playbook off the rule and onto the report

### cuz fotos work well, first one shows a report with no playbook, properly indicated, second one shows kba success as well as, remediate button working again
Screen Shot 2019-03-25 at 9 43 18 AM
Screen Shot 2019-03-25 at 9 42 40 AM

### DIFF
--- a/src/SmartComponents/Inventory/applications/advisor/Advisor.js
+++ b/src/SmartComponents/Inventory/applications/advisor/Advisor.js
@@ -30,7 +30,9 @@ class InventoryRuleList extends Component {
     }
 
     processRemediation (systemId, reports) {
-        const issues = reports.filter(r => r.rule.has_playbook).map(r => ({ id: `advisor:${r.rule.rule_id}`, description: r.rule.description }));
+        const issues = reports.filter(r => r.resolution.has_playbook).map(
+            r => ({ id: `advisor:${r.rule.rule_id}`, description: r.rule.description })
+        );
 
         return issues.length ?
             { issues, systems: [ systemId ]} :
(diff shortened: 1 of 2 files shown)


src/SmartComponents/Inventory/applications/advisor/Advisor.js
```diff
@@ -66,7 +68,7 @@ class InventoryRuleList extends Component {
 
     async fetchKbaDetails (kbaIds) {
         try {
-            const data = await fetch(`/rs/search?q=id:(${kbaIds})&fl=view_uri,id,publishedTitle`, { credentials: 'include' })
+            const data = await fetch(`https://access.redhat.com/rs/search?q=id:(${kbaIds})&fl=view_uri,id,publishedTitle`, { credentials: 'include' })
```
karelhala: Do we realy want to include `https://access.redhat.com` in here? I guess the best way would be to add this to spandx locally and to 3scale as well so once this is migrated to c.rh.com we are good to go with removing these rules from these and not by changing code.

AllenBW: If you gotta better way to do this, yes, please I am all ears, would you mind dropping in a commit @karelhala ?

karelhala: My bad, I didn't understood the reason why you want to include it there. Good to go!

PR #363, opened 2019-03-25
Allow passing raw html to Truncate

akofink:
For https://github.com/RedHatInsights/compliance-frontend/pull/103

### DIFF
--- a/src/PresentationalComponents/Truncate/Truncate.js
+++ b/src/PresentationalComponents/Truncate/Truncate.js
@@ -7,6 +7,8 @@ import { Button, Stack, StackItem } from '@patternfly/react-core';
 
 import './truncate.scss';
 
+import sanitizeHtml from 'sanitize-html';
+
 class Truncate extends React.Component {
 
     constructor(props) {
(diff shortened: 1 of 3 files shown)


src/PresentationalComponents/Truncate/Truncate.js
```diff
@@ -24,6 +24,10 @@ class Truncate extends React.Component {
         });
     }
 
+    dangerousHtml(text) {
+        return { __html: text };
```
ryelo: Do we want to sanitize this html before setting it?

If so, we may want to use something like: https://github.com/cure53/DOMPurify

akofink: Sounds like a good idea. I'll look into it. Thanks for the review!

akofink:
Updated with DOMPurify

package.json
```diff
@@ -22,6 +22,7 @@
     "c3": "^0.6.7",
     "classnames": "^2.2.5",
     "d3": "^5.7.0",
+    "dompurify": "^1.0",
```
karelhala: Advisor is using sanitize-html [1] please use it as well so we don't have two packages for same thing.

[1] https://github.com/RedHatInsights/insights-frontend-components/blob/master/src/SmartComponents/Inventory/applications/advisor/ExpandableRulesCard.js#L7

akofink:
Updated with sanitize-html

PR #365, opened 2019-03-26
Fixes RHICOMPL-86: SystemPolicyCard ref_id truncates

akofink:
image

image

https://projects.engineering.redhat.com/browse/RHICOMPL-86

### DIFF
--- a/src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
+++ b/src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { CheckCircleIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
+import Truncate from 'react-truncate';
 import {
     Card,
     CardBody,
@@ -11,10 +12,13 @@ import {
 import PropTypes from 'prop-types';
 
 class SystemPolicyCard extends React.Component {
-    complianceIcon = () => {
-        const { policy: { compliant }} = this.props;
+    constructor(policy) {
+        super(policy);
+        this.state = { refIdTruncated: <Truncate lines={ 1 }>{ policy.policy.ref_id }</Truncate>, ...policy };
+    }
 
-        return compliant ?
+    complianceIcon = () => {
+        return this.state.policy.compliant ?
             <div className='ins-c-policy-card ins-m-compliant'>
                 <CheckCircleIcon /> Compliant
             </div> :
@@ -23,9 +27,15 @@ class SystemPolicyCard extends React.Component {
             </div>;
     }
 
+    onMouseover = () => {
+        this.setState({ refIdTruncated: this.state.policy.ref_id });
+    }
+
+    onMouseout = () => {
+        this.setState({ refIdTruncated: <Truncate lines={ 1 }>{ this.state.policy.ref_id }</Truncate> });
+    }
+
     render() {
-        /* eslint-disable camelcase */
-        const { policy: { name, compliant, ref_id, last_scanned, rules_passed, rules_failed }} = this.props;
         return (
             <Card>
                 <CardBody>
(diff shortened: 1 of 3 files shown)


src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { CheckCircleIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
+import Truncate from 'react-truncate';
```
karelhala: Isn't this essentially the same as our long text truncate? [/src/PresentationalComponents/LongTextTooltip/LongTextTooltip.js](https://github.com/RedHatInsights/insights-frontend-components/blob/master/src/PresentationalComponents/LongTextTooltip/LongTextTooltip.js)

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
```diff
@@ -11,10 +12,16 @@ import {
 import PropTypes from 'prop-types';
 
 class SystemPolicyCard extends React.Component {
-    complianceIcon = () => {
-        const { policy: { compliant }} = this.props;
+    constructor(policy) {
+        super();
+        this.policy = policy.policy;
```
karelhala: Please do not assign to class attributes, but use state. With this approach our component will not be re-rendered once policy is changed.

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
akofink: This isn't a tooltip. Is that component misnamed or something?

akofink:
Updated to only use `this.state`

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
ryelo: I think @karelhala meant to link: https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/truncate.md

akofink: Is there a way to use auto length? How am I supposed to know how many characters? That depends on the width of the browser, right?

akofink:
What if it wraps instead of overflows? Is that okay?

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
```diff
@@ -11,10 +12,15 @@ import {
 import PropTypes from 'prop-types';
 
 class SystemPolicyCard extends React.Component {
-    complianceIcon = () => {
-        const { policy: { compliant }} = this.props;
+    constructor(policy) {
+        super();
```
karelhala: You should pass the policy to parent as well. `super(policy)`.

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
```diff
@@ -11,10 +12,15 @@ import {
...
+        this.state = { refIdTruncated: <Truncate lines={ 1 }>{ policy.policy.ref_id }</Truncate>, ...policy };
+        this.onMouseover = this.onMouseover.bind(this);
```
karelhala: No need to use binds.

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
```diff
@@ -23,29 +29,39 @@ class SystemPolicyCard extends React.Component {
             </div>;
     }
 
+    onMouseover() {
```
karelhala: You should use arrow function to automatically bind `this`
```JS
onMouseOver = () => {...}
```

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
```diff
@@ -23,29 +29,39 @@ class SystemPolicyCard extends React.Component {
...
+        this.setState({ refIdTruncated: this.state.policy.ref_id });
+    }
+
+    onMouseout() {
```
karelhala: You should use arrow function to automatically bind `this`
```JS
onMouseout= () => {...}
```

akofink:
Updated to remove binds and pass policy to `super()`. I also refactored a card in our frontend in a similar way: https://github.com/RedHatInsights/compliance-frontend/pull/113

karelhala:
Except for the truncate thing and @ryelo's request looks good! Thank you.

src/SmartComponents/Inventory/applications/compliance/SystemPolicyCard.js
ryelo: That's true. We may want to go back and revisit that issue with the other component. This should be fine for now.

akofink:
Updated to wrap the untruncated text:

image

PR #368, opened 2019-03-28
Add overriding DownloadButton props

akofink:
Set a custom toggle! Set custom downloadItems! :tada:


akofink:
I couldn't figure out how to use the method for `onToggle` in this component. :man_shrugging:

PR #369, opened 2019-03-29
Use inventory JS client instead of fetch

karelhala:


karelhala:
Tested in host inventory and as embeded both works. So feel free to review @ryelo @jiridostal

PR #373, opened 2019-04-04
switch remediations modal to new URL scheme

jharting:


src/SmartComponents/Remediations/utils.js
```diff
@@ -1,5 +1,16 @@
 import urijs from 'urijs';
 
+// Get the current group since we can be mounted at two urls
+export function getGroup () {
+    const pathName = window.location.pathname.split('/');
+
+    if (pathName[1] === 'beta') {
+        return pathName[2];
```
ryelo: Do we need to return both pathname[1] and [2] here?

PR #377, opened 2019-04-06
Add General info from system profile endpoint

karelhala:
This will update general information to newest mocks.

TODO:
* [x] 0/None records should not be clickable
* [x] Rename in actions should show dialog (new PR https://github.com/RedHatInsights/insights-frontend-components/pull/386)

Screenshot from 2019-04-11 11-32-35
Screenshot from 2019-04-11 11-32-46
Screenshot from 2019-04-10 16-18-49
Screenshot from 2019-04-10 16-18-56
Screenshot from 2019-04-10 16-19-11
Screenshot from 2019-04-10 16-19-28
Screenshot from 2019-04-10 16-19-36
Screenshot from 2019-04-10 16-19-43
Screenshot from 2019-04-10 16-19-50
Screenshot from 2019-04-11 18-07-51


karelhala:
ping @katierik can you review this from UX perspective? All clickable items lead to table with sortable items. Single cell tables looks kinda strange with table only, but I don't know which component would suit this better.

katierik:
The ["Data List" component](https://v2.patternfly.org/documentation/react/components/datalist) may be better than the table. It doesn't have the column headers. Unless there is more relevant content to put in there.
Thinking about consistency in what opens on the modal, I don't think its an issue to have a table open vs a list. 

Q: Is that "close" button required for our patterns? I posted on the pf forum to see if anyone had input. I lean towards not including it for simplicity https://forum.patternfly.org/t/read-only-modals/181

CC-ed this to Ben for review as well

bclarhk:
The text wrapping has to go, especially when there's plenty of horizontal room.
+1 to getting rid of the column headers but only when there is only one column of info.
I'm good with losing the Close button when the modal is info only; it's a cleaner pattern.
I'm double checking if the grey background is supposed to be there or not on the General info page. The mocks say no but I'll confirm.

karelhala:
@bclarhk @katierik Thanks for the review!

> The text wrapping has to go, especially when there's plenty of horizontal room.

Yeah, I'll have to look into it and see why the table wraps these columns.

> Column header

There were sadly no mocks for these dialogs and I went with table because it can show a lot of data in somewhat compact way (there can be a lot of packages for instance). Plus table gives you option to sort items, so I don't think it's good idea to remove the colum from there. However I would expect for these single column items to have some way of filtering.

> Close button

Yeah we can loose it, but do we want to? The way how modals are made is kinda unfortunate, you can't close modal by clicking away so I wanted to add another option to close the modal (x button at top is kinda small and hard to see)

> Background is supposed to be there or not

In Low-fi mocks there is gray background, in high-fi mocks it was missing so I assumed it should be there because other tabs has this background and it would look off to have it in one tab and not in the others.

bclarhk:
> There were sadly no mocks for these dialogs and I went with table because it can show a lot of data in somewhat compact way (there can be a lot of packages for instance). Plus table gives you option to sort items, so I don't think it's good idea to remove the colum from there. However I would expect for these single column items to have some way of filtering.

My fault for not having the Modal example(s) in there. Especially for the short timeframe we have, let's massage what's there into a cleaner look. Only show column headers when there's more than one column, and remove the sorting - but make everything sort by the first (or only) column. 
> 
> > Close button
> 
> Yeah we can loose it, but do we want to? The way how modals are made is kinda unfortunate, you can't close modal by clicking away so I wanted to add another option to close the modal (x button at top is kinda small and hard to see)

There should be a way to dismiss the modal by clicking outside of it. We've had to prevent it from happening in other projects. I'll track this down.
> 
> > Background is supposed to be there or not
> 
> In Low-fi mocks there is gray background, in high-fi mocks it was missing so I assumed it should be there because other tabs has this background and it would look off to have it in one tab and not in the others.

The gray background "Card view" is harder to read when there's so much text to scan. The Invision screen has been updated.

karelhala:
@bclarhk screenshots updated based on new proposed layout and such. Do you think you could give me example what to show if no data are returned? If the collection mechanism didn't collected any sysem profile data or server error happened when fetching system profile data.

karelhala:
@bclarhk I updated the icons for enabled/disabled as well. Sadly PF does not have that forbidden icon (circle with slash) so I used times icon as replacement.

karelhala:
@bclarhk I have updated storage modal with collapsible table, right now PF does not ship compact table with collapsible rows correctly (missing some styles for that).

src/SmartComponents/Inventory/applications/general-information/GeneralInformation.js
```diff
@@ -0,0 +1,118 @@
...
+    };
+
+    componentDidMount() {
+        this.props.loadSystemDetail(this.props.entity.id);
+    };
+
+    render() {
+        const { isModalOpen, modalTitle, cells, rows, expandable } = this.state;
+        return (
+            <Grid sm={ 6 } md={ 6 } lg={ 6 } gutter={ GutterSize.md }>
```
ryelo: You don't need to set these on each breakpoint. Also, we should probably have these each span 12 columns on small

src/SmartComponents/Inventory/applications/general-information/InfoTable.js
```diff
@@ -0,0 +1,92 @@
...
+            {
+                cells: [{ title: child }],
+                parent: key * 2
+            }
+        ]) : rows;
+        return (
+            <Fragment>
+                {
+                    cells.length !== 1 ? <Table
+                        aria-label="Compact Table"
```
ryelo: I'd change this to something more descriptive.

src/SmartComponents/Inventory/applications/general-information/InfrastructureCard.js
```diff
@@ -0,0 +1,73 @@
...
+                );
+            }
+        },
+        {
+            title: 'Inerfaces/NICs',
+            value: infrastructure.nics ? `${infrastructure.nics.length} NICs` : 0,
+            target: 'interfaces',
+            onClick: () => {
+                handleClick(
+                    'Inerfaces/NICs',
```
ryelo: spelling

src/SmartComponents/Inventory/applications/general-information/InfrastructureCard.js
```diff
@@ -0,0 +1,73 @@
...
+            target: 'ipv6',
+            onClick: () => {
+                handleClick(
+                    'IPv6',
+                    generalMapper(infrastructure.ipv6, 'IP address')
+                );
+            }
+        },
+        {
+            title: 'Inerfaces/NICs',
```
ryelo: spelling

src/SmartComponents/Inventory/applications/general-information/LoadingCard.js
```diff
@@ -0,0 +1,98 @@
...
+        onClick: PropTypes.func,
+        size: PropTypes.oneOf(Object.values(SkeletonSize))
+    }))
+};
+
+LoadingCard.defaultProps = {
+    isLoading: true,
+    items: [
+        {
+            title: 'bla',
```
ryelo: Can we just keep these at empty strings? like `title: ' '`

src/SmartComponents/Inventory/applications/general-information/SystemCard.js
```diff
@@ -0,0 +1,65 @@
...
+
+const SystemCard = ({ detailLoaded, entity, properties, handleClick }) => (<LoadingCard
+    title="System properties"
+    isLoading={ !detailLoaded }
+    items={ [
+        { title: 'Name', value: entity.display_name, size: 'md' },
+        { title: 'Number of CPUs', value: properties.cpuNumber },
+        { title: 'Sockets', value: properties.sockets },
+        { title: 'Cores per socket', value: properties.coresPerSocket },
+        { title: 'Ram', value: properties.ramSize },
```
ryelo: should this be capital? `RAM`

src/SmartComponents/Inventory/applications/general-information/GeneralInformation.js
karelhala: Yeah, good point. SM = 12 and MD = 6 is fine I guess.

src/SmartComponents/Inventory/applications/general-information/LoadingCard.js
karelhala: Yeah, these were testing strings. Changing it to empty array.

PR #376, opened 2019-04-05
Add empty states for Vulnerability

jiridostal:
* Adds action handler for Rejected actions
* Adds "errors" stanza to data structure
* Adds some of proposed Empty states and stores them in a separate file
image
image
image
image

PR #379, opened 2019-04-10
Vulnerability: Start using new filter component

jiridostal:
* Add new filters on cvvs_score, public_date, severity and status
* Add handling of empty state when filters are set
* Add undefined check to Filter component
Dependent on https://github.com/RedHatInsights/insights-frontend-components/pull/376 and https://github.com/RedHatInsights/insights-frontend-components/pull/378
image


karelhala:
Looks like PR on top of PR of https://github.com/RedHatInsights/insights-frontend-components/pull/376 we will wait once the empty state is OK.

PR #380, opened 2019-04-11
Inventory table redesign

karelhala:
Visually small changes
 * Used correct layout to show inventory actions
 * Two paginations, one on top the other at bottom
 * When filtering show empty message

Screenshot from 2019-04-11 15-09-38
Screenshot from 2019-04-11 16-00-28


karelhala:
@jiridostal this change will require some minor style changes to your filer, is that OK?

Screenshot from 2019-04-11 16-06-28

katierik:
> Visually small changes
> 
> * Used correct layout to show inventory actions
> * Two paginations, one on top the other at bottom
> * When filtering show empty message
> 
> Screenshot from 2019-04-11 15-09-38
> Screenshot from 2019-04-11 16-00-28

Well this is late, but the text styles on the empty state are a bit off. See the documentation: https://v2.patternfly.org/documentation/react/components/emptystate
The top header should be "pf-c-title pf-m-lg" and the supporting text should be "pf-c-empty-state__body"

AllenBW:
@katierik so this table is used in insights, and feedback there was to change "last sync" to "last seen", use relatvie time, and also throw the results count in the upper right (unsure how that works with pagination)... I just wanted to confirm that is all still work yah wanted done to this piece...
Screen Shot 2019-04-16 at 10 14 30 AM

karelhala:
> The top header should be "pf-c-title pf-m-lg" and the supporting text should be "pf-c-empty-state__body"

I would love to use PF react empty sate, but they do not allow empty state without icon. However good point with the classnames, I'll update it.

> "last sync" to "last seen" and use relatvie time

That will come in follow up PR, thanks for pointing this out.

> throw the results count in the upper right

I see this for the first time in mocks, however I think that the best way here would be to show per page dropdown option and pagination on bottom. Can we have some mocks variation to paginated inventory table with multiple buttons and no buttons?

PR #385, opened 2019-04-13
When fetching hosts with empty array do not dispatch error

karelhala:
@jiridostal this will remove temporary Empty table when fetching data in embedded inventory.

Basically dispatch empty response if we are trying to fetch hosts without list of hosts (filtered data, empty array passed as items to inventory).


jiridostal:
image
image
The first screenshot is what this PR introduces, and looks good! I've realized in the second screenshot that the "Status" column is suddenly empty, but I suspect more 62a824cf2afd1a50931d01dee9acd1074e54d705 than this PR.

karelhala:
@jiridostal updated the code, however now there is empty state flicker when you are using inv table. However that is caused by how you are using inventory, by default if you are fetching vuln systems you set items for inventory as empty array and once you fetch these items you pass them down. I guess the best way here would be if loading show spinner and once you get systems data show inventory table.

PR #387, opened 2019-04-16
Create empty states for insights detail

karelhala:
### Before
Screenshot from 2019-04-16 15-31-38

### After
Screenshot from 2019-04-16 15-24-25
Screenshot from 2019-04-16 15-29-12
Screenshot from 2019-04-16 15-30-14


src/SmartComponents/Inventory/applications/advisor/EmptyState.js
```diff
@@ -0,0 +1,34 @@
...
+import PropTypes from 'prop-types';
+
+const AdvisorEmptyState = ({ title, text, ...props }) => (<Bullseye { ...props }>
+    <EmptyState>
+        <EmptyStateIcon icon={ ChartSpikeIcon || CubesIcon } size="lg" />
+        <Title headingLevel="h5" size="lg">{ title }</Title>
+        <EmptyStateBody>
+            { text }
+        </EmptyStateBody>
+        <Button variant="primary">Try it free</Button>
```
AllenBW: should this go anywhere at the moment?

karelhala: Not really, we don't have correct links as of yet.

bsquizz:
What do we display for a system that has no Insights rule hits? I think that none of the 3 screens necessarily apply :)

You already have "gotten started with Insights", so screen 1 wouldn't be appropriate, and screen 3 wouldn't make much sense either because the system is in fact in inventory ...

In a case like this the screen should essentially display something like "No issues reported by Insights for this system at the moment."

karelhala:
@bsquizz good question, I just coppied over strings that were already used by advisor. But since we are updating these use cases we can rewrite these messages. Right now it is
1) System found, but has no issues - I think this one should be screen no. 3 and instead of this use case we should show something else
2) General error from API
3) No system found in insights DB

AllenBW:
@katierik wondering what the message(empty) state should be when there are no insight hits for a system? Only guidance I could find is this ss
Screen Shot 2019-05-07 at 7 12 53 AM

katierik:
@AllenBW This is the no insight hits for a system: https://redhat.invisionapp.com/share/V4NY3C6XKRW#/356911950_Empty_State_-_No_Insights



Also - The screenshots above don't match the mocks. All the mocks and use cases for empty states are here: https://docs.google.com/document/d/1iq1spvMCv3nc1epsv8KqtwuiRbH_6ZMmae53I4aoOWY/edit?usp=sharing

AllenBW:
@katierik the mocks do not appear to account for a system that has no reports (or no rule hits).. 🤔 any chance yah could provide direction to that affect?

katierik:
@AllenBW 
System detail, no rule hits: https://redhat.invisionapp.com/share/V4NY3C6XKRW#/356911950_Empty_State_-_No_Insights

AllenBW:
@karelhala  I wanta motion we close this for now, working on rebuilding the insights detail to be a table rather than a list, will be including table empty states n such in there 🤔

karelhala:
Sure thing, I'll close this so you can continue in followup PR

PR #388, opened 2019-04-16
Update inventory detail table styling to match mocks

karelhala:
### Before
Screenshot from 2019-04-10 16-19-11

### After
Screenshot from 2019-04-16 18-37-06

### DIFF
--- a/src/SmartComponents/Inventory/applications/general-information/general-information.scss
+++ b/src/SmartComponents/Inventory/applications/general-information/general-information.scss
@@ -15,7 +15,7 @@
 }
 
 .pf-c-modal-box .pf-c-table {
-    white-space:nowrap;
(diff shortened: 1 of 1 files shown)


src/SmartComponents/Inventory/applications/general-information/general-information.scss
```diff
@@ -15,7 +15,7 @@
 }
 
 .pf-c-modal-box .pf-c-table {
-    white-space:nowrap;
+    white-space: nowrap;
```
ryelo: Is this going to cause issues on mobile? I think setting the min-width a larger, percentage based value would be better unless the mocks specifically call for these smaller modals. If that's the case, this should be fine as long as wrapping is not ever needed on smaller screens.

PR #389, opened 2019-04-16
Tweak the wording when there's no rule hits on a system

bsquizz:
I think the current wording here is a little confusing, it almost reads like an error message indicating that "we have failed to fetch the data" :) But in reality, this message will be seen for a system that has no rule hits.


bsquizz:
@AllenBW thoughts?

AllenBW:
@bsquizz https://github.com/RedHatInsights/insights-frontend-components/pull/387/

🤤

karelhala:
Let's merge this one and update #387 accordingly

PR #406, opened 2019-04-29
Regex replaces all instances, template literal (string) only 1st (ul instance)

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-705

### now we're replacing all instances
Screen Shot 2019-04-29 at 8 52 59 AM
Screen Shot 2019-04-29 at 8 52 49 AM

### DIFF
--- a/src/SmartComponents/Inventory/applications/advisor/ExpandableRulesCard.js
+++ b/src/SmartComponents/Inventory/applications/advisor/ExpandableRulesCard.js
@@ -52,7 +52,7 @@ class ExpandableRulesCard extends React.Component {
             const compiledMd = marked(sanitizeHtml(compiledDot, sanitizeOptions));
 
             return <div dangerouslySetInnerHTML={ { __html: compiledMd
-            .replace(`<ul>`, `<ul class="pf-c-list">`)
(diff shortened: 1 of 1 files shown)


AllenBW:
cc @AparnaKarve

AparnaKarve:
@AllenBW Thank you! You nailed it - the bullets are looking perfect now!!

Just wondering how does the compiled pydata block look like...
Does it have marked-down line-breaks after the line `<hr>`?... since that is what we are sending..

As a workaround, I was thinking of explicitly specifying `<br>` after the line(`<hr>`) since the markdown line breaks are not working in the UI. 
But before I do that, just wanted to know if it was doable through the insights-frontend-components route.
Thanks!

AllenBW:
@AparnaKarve found this neat thing about linebreaks in md...https://gist.github.com/shaunlebron/746476e6e7a4d698b373

we don't do anything to <hr> but don't _explicitly_ support html so if it works 👍 if not 🤷‍♂

AllenBW:
yah might want to try doing an hr in md https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#horizontal-rule

PR #413, opened 2019-04-30
Ensures rule hits remain whole when not consumed by insights

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1117

tl;dr splice mutates the original array. 👻 🐍 👎 

if  you do something like `reports.splice(-1, 1);` (which is what happens when we don't find an active report), you lose the last entry from the list, which is what we see in inventory app 😭

### DIFF
--- a/src/SmartComponents/Inventory/applications/advisor/Advisor.js
+++ b/src/SmartComponents/Inventory/applications/advisor/Advisor.js
@@ -74,7 +74,7 @@ class InventoryRuleList extends Component {
     }
 
     sortActiveReports (activeReports) {
-        const reports = Object.assign({}, activeReports);
+        const reports = { ...activeReports };
         const activeRuleIndex = activeReports.findIndex(report => report.rule.rule_id === this.props.match.params.id);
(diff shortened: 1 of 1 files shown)


AllenBW:
cc @dkuc

src/SmartComponents/Inventory/applications/advisor/Advisor.js
```diff
@@ -74,7 +74,7 @@ class InventoryRuleList extends Component {
     }
 
     sortActiveReports (activeReports) {
-        const reports = activeReports;
+        const reports = Object.assign({}, activeReports);
```
karelhala: Isn't spread operator better here? `{...activeReports}`

AllenBW: OOOOO seee i had this same conversation with @dkuc https://stackoverflow.com/questions/32925460/object-spread-vs-object-assign had some thoughts...  didn't ultimately come to any conclusion...

PR #414, opened 2019-04-30
Adapt to new Remediations API in Compliance

dLobatog:


src/SmartComponents/Inventory/applications/compliance/ComplianceRemediationButton.js
```diff
@@ -2,60 +2,91 @@ import React from 'react';
...
 class ComplianceRemediationButton extends React.Component {
     constructor(props) {
         super(props);
     }
 
     /* eslint-disable camelcase */
-    formatRule = ({ title, ref_id }, system) => ({
-        id: `compliance:${ref_id}`,
+    formatRule = ({ title, ref_id }, profile, system) => ({
+        id: `ssg:rhel7|${profile}|${ref_id}`,
```
jharting: @dLobatog this should not be hardcoded to rhel7, right?

src/SmartComponents/Inventory/applications/compliance/ComplianceRemediationButton.js
```diff
@@ -2,60 +2,91 @@ import React from 'react';
...
 
     render() {
-        const { disableRemediations } = this.props;
+        const { allSystems, selectedRules } = this.props;
+
         return (
             <React.Fragment>
                 <RemediationButton
-                    isDisabled={ disableRemediations || this.dataProvider().issues.length === 0 }
+                    isDisabled={ (allSystems.length === 0 || allSystems[0].rule_objects_failed.length === 0) &&
```
jharting: This condition looks suspicious - why checking the first system? Should not there be disjunction instead of conjunction?

dLobatog: I basically copied over the ComplianceRemediationButton in `compliance-frontend` to here. In here, the button is used at a page where there's only one host

dLobatog:
Here's a quick demo:

Peek 2019-05-01 20-53

src/SmartComponents/Inventory/applications/compliance/ComplianceRemediationButton.js
dLobatog: I am hardcoding it as I'm checking all of the inventory facts, yet none of them appear to give me any clear answer regarding the RHEL version, `os.release`, etc.. don't seem to contain that. How are you checking that? Would it be OK to default to rhel7 for the moment?

I have looked into it and I think we can infer that info from the OpenSCAP report, but I would like to get the Remediate With Ansible button to work ASAP (as marketing/CEE are trying to use it) before we fetch the OS release from the report. 

If you know of any way of getting it from the inventory, that'd be great too. Here's an example of an inventory host: 

https://gist.github.com/c8a41ce298f0a17c456b4429764644c6

PR #427, opened 2019-05-07
Add 'credentials' to apolloclient for Firefox <= 60

dLobatog:


karelhala:
Good to go?

dLobatog:
Thanks @karelhala  - this fixed the issue in compliance-frontend, we should add this to inventory-frontend

PR #428, opened 2019-05-07
Replaces the kbafetch with static `node/${report.rule.node_id}`

AllenBW:
Similar to https://github.com/RedHatInsights/insights-advisor-frontend/pull/316 but what we lose is article title on the link... what we save is making yet another external call... need ux input on this, it says "articles" but we only ever display an article, (this is a carry over from classic)

### what it looks like
Screen Shot 2019-05-07 at 1 30 07 PM

(kinda) fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1195


AllenBW:
@karelhala we need to getttttttttt ux approval of for this i believe...

AllenBW:
closing cuz BU said nope.

PR #435, opened 2019-05-13
adding width to pagination input <?.?.x>

ryelo:
Temporary fix to allow for 5 numbers


AllenBW:
If only this pr could hit so hard it changed pf4 😭

